PR #2, opened 2023-11-09
Add AAP Job resource

abikouo:
This new resource should allow creating Job on AAP using a Job template id and optionally wait for it to be completed

Example configuration

```

terraform {
  required_providers {
    aap = {
      source = "ansible/aap"
    }
  }
}

provider "aap" {
  host     = "https://localhost:8043"
  username = "ansible"
  password = "test123!"
  insecure_skip_verify = true
}

variable "extra_vars" {
  type = any
}

resource "aap_job" "sample" {
  job_template_id   = 10
  inventory_id = 1
  wait_for_completion = false
  wait_duration = 10
  extra_vars = jsonencode(var.extra_vars)
}

output "job_launch_url" {
  value = aap_job.sample.job_url
}
```

with a `terraform.tfvars.json` configure like this

```
{
    "extra_vars": {
        "simple": "testing",
        "simple_bool": false,
        "simple_object": {
            "key1": "value1",
            "key2": "value2"
        }
    }
}
```


internal/provider/job_resource.go
```diff
@@ -0,0 +1,518 @@
...
+			return
+		}
+		data.Status = types.StringValue(job_status)
+	}
+
+	// Save updated data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r JobResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
```
jillr, line 367: I don't think we should be deleting jobs by default; many users may want to keep their job launch history to be able audit it.  The user running the job may not even have access to delete things.  What prompted the decision to delete jobs in Controller?

hakbailey: All four CRUD operations are [expected for resources](https://developer.hashicorp.com/terraform/plugin/framework/resources) in the terraform provider framework. I was chatting about this with Mike the other day, but IMO launching an AAP job is similar to invoking a lambda function in AWS and I'm wondering if we want our provider to handle it in a similar way? I notice the AWS provider allows you to invoke a lambda either as [a resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_invocation) or as [a data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/lambda_invocation), which is interesting.

abikouo: @jillr as mentioned by @hakbailey we need all CRUD operation to be defined for a resource. Now we can discuss about the delete operation whether it should simply cancel the job or fully delete it. Also note that if the Job was launched by the user they should be able to delete it too.

internal/provider/job_resource.go
```diff
@@ -0,0 +1,518 @@
...
+			"status": schema.StringAttribute{
+				Computed: true,
+			},
+			"wait_for_completion": schema.BoolAttribute{
+				Optional: true,
+			},
+			"wait_duration": schema.Int64Attribute{
+				Optional: true,
+			},
+			"extra_vars": schema.StringAttribute{
```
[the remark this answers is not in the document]
abikouo: I tried to define it as map, but it will be only map of strings knowing that ansible variable can be dict/string/list... This is definitely the recommended way to deal with this attribute as this is threat on the provider side but directly passed to the server. More info [here](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#dynamic-types-the-any-constraint)

internal/provider/job_resource.go
```diff
@@ -0,0 +1,518 @@
...
+				ElementType: types.StringType,
+				Computed:    true,
+			},
+		},
+	}
+}
+
+// jobResourceModel maps the resource schema data.
+type jobResourceModel struct {
+	Id                   types.Int64  `tfsdk:"job_template_id"`
```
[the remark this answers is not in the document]
abikouo: yes sure

examples/resources/job/main.tf
```diff
@@ -0,0 +1,30 @@
...
+variable "extra_vars" {
+  type = any
+}
+
+resource "aap_job" "sample" {
+  job_template_id   = 10
+  inventory_id = 1
+  wait_for_completion = false
+  wait_duration = 10
+  extra_vars = jsonencode(var.extra_vars)
```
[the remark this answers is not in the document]
abikouo: done

internal/provider/client.go
```diff
@@ -63,18 +87,46 @@ func (c *AAPClient) GetHosts(stateId string) (*AnsibleHostList, error) {
...
+
+func (c *AAPClient) doRequest(method string, path string) (int, []byte, error) {
+
+	req, _ := http.NewRequest(method, c.computeURLPath(path), nil)
+	return c.SendRequest(req)
+}
+
+func (c *AAPClient) GetHosts(stateId string) (*AnsibleHostList, error) {
+
+	hostURL := c.HostURL
```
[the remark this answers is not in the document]
abikouo: yes it is duplicated!! 
Removed

internal/provider/client.go
```diff
@@ -41,14 +46,33 @@ func NewClient(host string, username *string, password *string, insecure_skip_ve
...
 }
 
-func (c *AAPClient) GetHosts(stateId string) (*AnsibleHostList, error) {
-
-	hostURL := c.HostURL
-	if !strings.HasSuffix(hostURL, "/") {
-		hostURL = hostURL + "/"
+func (c *AAPClient) computeURLPath(path string) string {
+	fullPath := c.HostURL
+	if !strings.HasSuffix(fullPath, "/") {
```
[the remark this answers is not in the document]
abikouo: removed. Thanks

internal/provider/client.go
```diff
@@ -41,14 +46,33 @@ func NewClient(host string, username *string, password *string, insecure_skip_ve
...
+		fullPath = fullPath + path
 	}
+	if !strings.HasSuffix(fullPath, "/") {
+		fullPath = fullPath + "/"
+	}
+	return fullPath
+}
+
+func (c *AAPClient) SendRequest(req *http.Request) (int, []byte, error) {
+	// 	// add query params into request
```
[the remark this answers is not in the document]
abikouo: No reason to keep it.
Removed.

internal/provider/job_resource.go
gravesm: While the CRUD operations have to be there for the Terraform data model, delete and update don't actually have to do anything on the AAP side. I'm not sure that they should, either. Our goal isn't to build a Terraform resource that fully implements the AAP API, our goal is to allow Terraform users to connect changes in an inventory to launching a job template. This is really the primary use case. I think we want to emphasize the lifecycle connection between Instance --> Inventory --> Job.

As @hakbailey suggested, the way the AWS provider handles the lambda invocation is probably a good model to look at. I'm not sure that using a data source is the right approach as this would launch a job template for every apply. Using a resource gives the user the flexibility to decide whether the job template should be launched only on create, or when some other arbitrary property has changed. For example, you might add a trigger that looks at the contents of an inventory, so the job template gets launched whenever hosts are added/removed from the inventory.

jillr: Thanks y'all.  I think we should avoid deleting jobs on the AAP side as I expect a lot of AAp users would be surprised by that.

abikouo: I just want to summarize for everyone to be aligned, 

***Current implementation***

- _Update_: a new job will be triggered if one of the following has changed: job_template_id, ee_id, inventory_id or extra_vars, this is internal TF logic
- _Delete_: cancel running job and delete it from AAP 

***What we are expecting***

- _Update_: ? I am not clear about what you are expecting
- _Delete_: Cancel running Job if not finished

@jillr @gravesm

gravesm: When an update is required, launch a new job. An update should be triggered when the inventory_id, job_template_id or extra_vars are changed. I would remove the ee_id and wait properties from the resource. Our goal, for now at least, should be to provide a minimally useful mechanism for spawning a job template. The expectation is still that the user is going to be interacting with the AAP UI to configure job templates and check the status of jobs.

The resource should also have a trigger property that accepts an arbitrary set of key/values. This can be used to trigger an update of the resource based on whatever criteria the user wants.

The resource should not cancel or delete an AAP job at all. Deleting just removes the resource from the TF state.

internal/provider/job_resource.go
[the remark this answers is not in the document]
hakbailey: Got it, makes sense. Thanks!

internal/provider/job_resource.go
```diff
@@ -0,0 +1,321 @@
...
+	return validator.Description(ctx)
+}
+
+// Validate performs the validation.
+func (v ansibleVarsValidator) ValidateString(ctx context.Context, request validator.StringRequest, response *validator.StringResponse) {
+	if request.ConfigValue.IsNull() || request.ConfigValue.IsUnknown() {
+		return
+	}
+	if !json.Valid([]byte(request.ConfigValue.ValueString())) {
+		response.Diagnostics.Append(validatordiag.InvalidAttributeValueLengthDiagnostic(
```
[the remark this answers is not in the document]
abikouo: updated

internal/provider/job_resource.go
```diff
@@ -0,0 +1,321 @@
...
+type JobResource struct {
+	client *AAPClient
+}
+
+// Metadata returns the resource type name.
+func (r *JobResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_job"
+}
+
+var _ validator.String = ansibleVarsValidator{}
```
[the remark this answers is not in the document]
abikouo: yes, this is a way to validate that the `ansibleVarsValidator` struct is implementing required functions

internal/provider/client.go
```diff
@@ -63,18 +74,41 @@ func (c *AAPClient) GetHosts(stateId string) (*AnsibleHostList, error) {
...
 	defer resp.Body.Close()
 
 	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return -1, []byte{}, err
+	}
+	return resp.StatusCode, body, nil
+}
+
+func (c *AAPClient) doRequestWithBody(method string, path string, req_data *bytes.Reader) (int, []byte, error) {
```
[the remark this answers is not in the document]
abikouo: I tried using one function with `nil` but was not working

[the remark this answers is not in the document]
gravesm: What do you mean by not working? It seems to compile without issues. I'd also suggest using an `io.Reader` instead of a `*bytes.Reader` here and elsewhere. In general, if there's an interface you can use, it's better to use that than a specific implementation of it.

internal/provider/job_resource.go
```diff
@@ -0,0 +1,321 @@
...
+			return
+		}
+	}
+
+	// Save updated data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r JobResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var data *jobResourceModel
```
gravesm, line 276: I'm guessing this is not doing what you want. The default value for a pointer is `nil`. If you want a pointer to a default `jobResourceModel` then use `data := &jobResourceModel{}`. I would suggest just creating the struct and when you need a pointer to it later, just get it then, similar to how you have done in the `Delete` method.

internal/provider/job_resource.go
```diff
@@ -0,0 +1,321 @@
...
+
+	// Read Terraform plan data into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
```
gravesm, line 279: `data` is already a pointer. You are passing a pointer to a pointer here.

internal/provider/job_resource.go
abikouo: done, thanks

internal/provider/client.go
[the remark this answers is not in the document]
abikouo: done

internal/provider/job_resource.go
```diff
@@ -0,0 +1,334 @@
...
+		)
+		return
+	}
+
+	// Save updated data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r *JobResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("job_template_id"), req, resp)
```
hakbailey, line 333: I don't think this will work because `ImportStatePassthroughID` requires that the state attribute accept a string but `"job_template_id"` is int64. Do we need to support state import for a job resource?

internal/provider/client.go
```diff
@@ -2,123 +2,73 @@ package provider
...
-		InsecureSkipVerify: insecure_skip_verify,
+		HostURL:  hostURL,
+		Username: username,
+		Password: password,
+	}
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure_skip_verify},
 	}
+	// Initialize http client using default timeout = 5sec
```
hakbailey, line 40: Where is the default timeout set to 5? It looks like the provider just uses a 0 value if nothing is passed?

internal/provider/job_resource.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	var http_code int
+	var body []byte
+	post_url := "/api/v2/job_templates/" + data.GetTemplateId() + "/launch/"
+	if req_data != nil {
+		http_code, body, err = r.client.doRequest("POST", post_url, req_data)
+	} else {
+		http_code, body, err = r.client.doRequest("POST", post_url, nil)
+	}
```
hakbailey, line 234: I think this if/else block can be replaced with just 
`http_code, body, err = r.client.doRequest("POST", post_url, req_data)`

internal/provider/client.go
abikouo: updated

internal/provider/job_resource.go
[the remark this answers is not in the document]
gravesm: I feel like this would be better off as a custom [normalized jsontype](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-jsontypes@v0.1.0/jsontypes#Normalized.Type). This way we aren't reliant on a user having used `jsonencode()`.

internal/provider/client.go
```diff
@@ -1,124 +1,78 @@
...
 }
 
-// NewClient -
-func NewClient(host string, username *string, password *string, insecure_skip_verify bool) (*AAPClient, error) {
+// NewClient - create new AAPClient instance
+func NewClient(host string, username *string, password *string, insecureSkipVerify bool, timeout int64) (*AAPClient, error) {
+	hostURL := host
+	if !strings.HasSuffix(hostURL, "/") {
+		hostURL += "/"
+	}
```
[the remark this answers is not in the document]
abikouo: done

internal/provider/client.go
```diff
@@ -1,124 +1,78 @@
...
 	}
+	return fullPath
+}
 
-	req, _ := http.NewRequest("GET", hostURL+"api/v2/state/"+stateId+"/", nil)
+func (c *AAPClient) doRequest(method string, path string, data io.Reader) (int, []byte, error) {
+	ctx := context.Background()
+	req, err := http.NewRequestWithContext(ctx, method, c.computeURLPath(path), data)
+	if err != nil {
+		return -1, []byte{}, err
```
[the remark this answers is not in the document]
abikouo: updated

internal/provider/provider.go
```diff
@@ -135,61 +125,43 @@ func (p *aapProvider) Configure(ctx context.Context, req provider.ConfigureReque
...
-		username = config.Username.ValueString()
-	}
-
-	if !config.Password.IsNull() {
-		password = config.Password.ValueString()
-	}
-
-	if !config.InsecureSkipVerify.IsNull() {
-		insecure_skip_verify = config.InsecureSkipVerify.ValueBool()
+	rawTimeout := os.Getenv("AAP_TIMEOUT")
```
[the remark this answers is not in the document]
abikouo: fixed

internal/provider/job_resource.go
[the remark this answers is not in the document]
abikouo: updated

gravesm (commented):
Can you remove the integration testing from this PR, please? Also, the job resource is still missing the manual trigger functionality that was discussed in a previous comment.

internal/provider/job_resource.go
```diff
@@ -0,0 +1,333 @@
...
+	return !value.IsNull() && !value.IsUnknown()
+}
+
+func (d *jobResourceModel) CreateRequestBody() (io.Reader, diag.Diagnostics) {
+	body := make(map[string]interface{})
+
+	// Extra vars
+	var diags diag.Diagnostics
+	if IsValueProvided(d.ExtraVars) {
+		var extraVars map[string]interface{}
```
hakbailey, line 179: Is it necessary to unmarshal this here?

internal/provider/job_resource.go
```diff
@@ -0,0 +1,333 @@
...
+	diags.Append(diagCreateReq...)
+	if diags.HasError() {
+		return diags
+	}
+
+	var postURL = "/api/v2/job_templates/" + data.GetTemplateID() + "/launch/"
+	resp, body, err := r.client.doRequest(http.MethodPost, postURL, reqData)
+
+	if err != nil {
+		diags.AddError("Body JSON Marshal Error", err.Error())
```
hakbailey, line 236: I don't think this will be a JSON Marshal error?

internal/provider/job_resource.go
```diff
@@ -0,0 +1,333 @@
...
+		return diags
+	}
+	if resp == nil {
+		diags.AddError("Http response Error", "no http response from server")
+		return diags
+	}
+	if resp.StatusCode != http.StatusCreated {
+		diags.AddError("Unexpecte Http Status code",
+			fmt.Sprintf("expected (%d) got (%d)", http.StatusCreated, resp.StatusCode))
```
hakbailey, line 245: It might be useful to also include the response here since it could include more detailed information than just the status code.

Makefile
```diff
@@ -0,0 +1,25 @@
+.PHONY: build test lint
+
+default: build
+
+build:
```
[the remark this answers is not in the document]
abikouo: That' strange it is working locally for me

Makefile
```diff
@@ -0,0 +1,25 @@
...
+lint:
+	@echo "==> Checking source code against linters..."
+	golangci-lint run -v ./...
+
+test:
+	@echo "==> Running unit tests..."
+	go test -v ./...
+
+testacc:
+	touch $(PWD)/.testacc_configure.sh
```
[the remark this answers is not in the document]
abikouo: Mikes asked to removed acceptance tests, so this is not relevant anymore.
Thanks

internal/provider/provider.go
```diff
@@ -58,10 +59,35 @@ func (p *aapProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp *
 			"insecure_skip_verify": schema.BoolAttribute{
 				Optional: true,
 			},
+			"timeout": schema.Int64Attribute{
+				Optional: true,
+				Description: "Timeout specifies a time limit for requests made to the AAP server." +
+					" A Timeout of zero means no timeout.",
```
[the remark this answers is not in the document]
abikouo: done

internal/provider/job_resource.go
```diff
@@ -0,0 +1,333 @@
...
+		response.Diagnostics.Append(validatordiag.InvalidAttributeValueDiagnostic(
+			request.Path,
+			v.Description(ctx),
+			fmt.Sprintf("Invalid JSON string => [%s]", request.ConfigValue.ValueString()),
+		))
+		return
+	}
+}
+
+func AnsibleJSONVarsValidator() validator.String {
```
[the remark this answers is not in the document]
abikouo: I just missed to remove it. Removed.

internal/provider/job_resource.go
abikouo: I was wrong, this is a client request error

internal/provider/job_resource.go
abikouo: it is not sure we always have a response in this case and adding it could make the log pretty heavy

internal/provider/job_resource.go
abikouo: yes it is. TF stores only a raw value which is not JSON-parsed, this can generate errors

hakbailey: But doesn't the normalized JSON type validator ensure that it's a valid JSON string?

abikouo: The normalized JSON type ensures that it is a valid JSON string. 
We can remove the unmarshall here, but this means the user should not escape quote into the JSON definition, the following code will be replaced by 

```golang
body["extra_vars"] = d.ExtraVars.ValueString()
```

and the main.tf will look like
```terraform
resource "aap_job" "sample" {
 ... 
 extra_vars = file("./vars.json")
  ...
}
```

with `vars.json`

```json
{ "some": "variable"}
```

In this case if the user configuration is like the following below, it will not work as expected

```terraform
resource "aap_job" "sample" {
 ... 
 extra_vars = "{ \"some\": \"variable\"}"
  ...
}
```

adding unmarshall here will resolve the issue

abikouo:
> Can you remove the integration testing from this PR, please? Also, the job resource is still missing the manual trigger functionality that was discussed in a previous comment.

The manual trigger has been added and the acceptance tests removed

hakbailey (approved):

internal/provider/job_resource_test.go
```diff
@@ -0,0 +1,622 @@
...
+}
+
+func TestCreateRequestBody(t *testing.T) {
+	testTable := []struct {
+		name     string
+		input    jobResourceModel
+		expected []byte
+	}{
+		{
+			name: "unknown fields",
```
GomathiselviS, line 112: What is the difference between this test and [this one](https://github.com/ansible/terraform-provider-aap/pull/2/files#diff-9783f57a592d7669379e472320b72831a01eec44cf40760d6a519357cfee72afR120) ?

abikouo: @GomathiselviS the link is pointing the #diff of the PR, please specify the file and the line or the test case name

GomathiselviS: Sorry about it. I am ltrying to understand the difference between 2 testcases - https://github.com/ansible/terraform-provider-aap/blob/68b6369c3b314fed66279ac93dfe85f5bf1116af/internal/provider/job_resource_test.go#L112 and https://github.com/ansible/terraform-provider-aap/blob/68b6369c3b314fed66279ac93dfe85f5bf1116af/internal/provider/job_resource_test.go#L120

abikouo: @GomathiselviS the Terraform `attr.value` can be `null` (empty value) or `unknown` (unset), these 2 test cases are different in the sense that they are testing these different cases

internal/provider/job_resource.go
```diff
@@ -0,0 +1,315 @@
...
+	body := make(map[string]interface{})
+	var diags diag.Diagnostics
+
+	// Triggers
+	if IsValueProvided(d.Triggers) {
+		// Set manual triggers
+		for key, value := range d.Triggers.Elements() {
+			body[key] = value.(types.String).ValueString()
+		}
+	}
```
gravesm, line 152: Triggers should not be added to the HTTP request. They are just there as a way to force TF to update the resource. The reason this is needed is because of the nature of the job resource. A user may often want to launch a new job using the same data. TF is only going to do that if something has changed with the job resource definition, though. It's possible to do this with `terraform taint`, but the `triggers` attribute is being included so that the user can do the same by tying the new launch to whatever TF values they want.

internal/provider/job_resource.go
```diff
@@ -0,0 +1,315 @@
...
+	if err != nil {
+		diags.AddError("client request error", err.Error())
+		return diags
+	}
+	if resp == nil {
+		diags.AddError("Http response Error", "no http response from server")
+		return diags
+	}
+	if resp.StatusCode != http.StatusCreated {
+		diags.AddError("Unexpecte Http Status code",
```
gravesm, line 226:
```suggestion
		diags.AddError("Unexpected Http Status code",
```

internal/provider/job_resource.go
abikouo: @gravesm it looks strange the way we want to use this field however I will update accordingly.
Is it ok if the field is defined as a string?

gravesm: It should be a map. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_invocation#triggers for an example of the pattern in use elsewhere in the TF ecosystem.

abikouo: updated

gravesm (approved):

PR #3, opened 2023-12-06
Add Group Resource

GomathiselviS:
This PR adds group resource. Refer https://issues.redhat.com/browse/ACA-1158

```
terraform {
  required_providers {
    aap = {
      source = "ansible/aap"
    }
  }
}

provider "aap" {
  host     = "https://controller.xxx.xyz/"
  username = "xxxx"
  password = "xxxx"
  insecure_skip_verify = true
}

resource "aap_group" "sample" {
  inventory_id = 1
  name = "tf_group" 
  variables = jsonencode({"ansible_network_os": "ios"})
}
```


internal/provider/group_resource.go
```diff
@@ -0,0 +1,366 @@
...
+        if IsValueProvided(d.Variables) {
+                // vars := make(map[string]interface{})
+                // vars_data := []byte(d.Variables)
+                // var replacer = strings.NewReplacer("\r", "", "\n", "")
+                // vars_data := replacer.Replace(d.Variables)
+                // diags.Append(d.Variables.Unmarshal(&vars)...)
+		// if diags.HasError() {
+		//	return nil, diags
+		// }
+		body["variables"] = d.Variables.ValueString()
```
gravesm, line 127: I think this needs to be unmarshalled, otherwise when you marshal the body later it will end up as just an escaped json string.

internal/provider/group_resource.go
```diff
@@ -0,0 +1,366 @@
...
+		return nil, diags
+	}
+	return bytes.NewReader(json_raw), diags
+}
+
+func (d *groupResourceModel) ParseHttpResponse(body []byte) error {
+	/* Unmarshal the json string */
+	result := make(map[string]interface{})
+        var replacer = strings.NewReplacer("\r", "", "\n", "")
+	data := replacer.Replace(string(body))
```
gravesm, line 150: This shouldn't be necessary.

internal/provider/group_resource.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+func (d *groupResourceModel) GetURL() string {
+        if !d.URL.IsNull() && !d.URL.IsUnknown() {
+		return d.URL.ValueString()
+	}
+	return ""
+}
+
+func GetFunctionName(caller int) string {
```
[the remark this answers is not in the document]
GomathiselviS: Although this function returns the function name, the function name itself is meaningful and helps users identify the action where the error occurred. It was beneficial when distinguishing where the exception was thrown, either during the group reading or updating process. For instance, it would display something similar to "ReadGroup: Json Marshal Error." Do you believe this feature is still unnecessary?

internal/provider/group_resource.go
GomathiselviS: Upon unmarshalling, the request is not accepted. As far as I understand, the expected format for variables is a string.

internal/provider/group_resource.go
```diff
@@ -0,0 +1,366 @@
...
+// Metadata returns the resource type name.
+func (r *groupResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+    resp.TypeName = req.ProviderTypeName + "_group"
+}
+
+// Schema defines the schema for the group resource.
+func (r *groupResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"id": schema.Int64Attribute{
```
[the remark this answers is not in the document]
GomathiselviS: The API output does not include this value. I will remove this field.

internal/provider/group_resource_test.go
```diff
@@ -0,0 +1,229 @@
...
+			InventoryId: basetypes.NewInt64Value(1),
+			Name:        types.StringValue("group1"),
+			URL:         types.StringValue("/api/v2/groups/24/"),
+		}
+		body := []byte(`{"inventory": 1, "name": "group1", "url": "/api/v2/groups/24/"}`)
+		result, diags := g.CreateRequestBody()
+		if diags.HasError() {
+			t.Fatal(diags.Errors())
+		}
+		test, err := DeepEqualJSONByte(body, result)
```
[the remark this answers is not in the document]
gravesm: I'm guessing the reason for this is because two JSON objects may be functionally equivalent, but may serialize to a different byte sequence because of key ordering. If you can guarantee the order of keys, you could just do a simple byte comparison, but it's probably not worth it. It's easier to just unmarshal and then compare the resulting structs.

internal/provider/group_resource.go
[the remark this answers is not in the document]
gravesm: Yes, we should remove it. The internal function name shouldn't be part of the user facing error message. The context in which TF is being run will already tell the user if the error happened while creating, deleting, etc.

internal/provider/group_resource_test.go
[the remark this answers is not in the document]
gravesm: I just noticed you added testify in this PR. If we're using testify, there's already a function that compares JSON strings: https://pkg.go.dev/github.com/stretchr/testify/require?utm_source=godoc#JSONEq

internal/provider/group_resource.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+type GroupResourceModelInterface interface {
+	ParseHttpResponse(body []byte) error
+	CreateRequestBody() ([]byte, diag.Diagnostics)
+	GetURL() string
+}
+
+// groupResource is the resource implementation.
+type groupResource struct {
```
gravesm, line 39:
```suggestion
type GroupResource struct {
```
I would make this exportable. I'm not sure there's a good reason not to, and for consistency, the job resource in the other PR is exportable.

internal/provider/group_resource.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource = &groupResource{}
```
abikouo, line 24:
```suggestion
	_ resource.Resource = &groupResource{}
	_ resource.ResourceWithConfigure = &groupResource{}
```

The groupResource defines `Configure` method, you need to add this to validate he struct

abikouo (changes requested):

internal/provider/group_resource.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	return diags
+}
+
+func (r GroupResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var data GroupResourceModel
+	var data_with_URL GroupResourceModel
+
+	// Read Terraform plan and state data into the model
+	// The URL is generated once the group is created. To update the correct group, we retrieve the state data and append the URL from the state data to the plan data.
```
hakbailey, line 293: I don't think this is necessary. With the `UseStateForUnknown` planmodifier, the plan should include the URL from state.

internal/provider/group_resource_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+func TestParseHttpResponse(t *testing.T) {
+	t.Run("Basic Test", func(t *testing.T) {
+		g := GroupResourceModel{
+			Name: types.StringValue("group1"),
+			URL:  types.StringValue("/api/v2/groups/24/"),
+		}
+		body := []byte(`{"name": "group1", "url": "/api/v2/groups/24/", "description": ""}`)
+		err := g.ParseHttpResponse(body)
+		assert.NoError(t, err)
```
hakbailey, line 27: I think we probably also want to test the values of the `GroupResourceModel` here to be sure they have been updated correctly. You can start with it empty to ensure all values are added (e.g., `g := GroupResourceModel{}`).

abikouo (commented):
LGTM, Just lint issues to fix

GomathiselviS:
> LGTM, Just lint issues to fix

Hi @abikouo Can you help me fix this [lint failure](https://github.com/ansible/terraform-provider-aap/actions/runs/7547915044/job/20548883411?pr=3#step:4:31). The failure happens in the job_resource_test.go

gravesm:
> Hi @abikouo Can you help me fix this [lint failure](https://github.com/ansible/terraform-provider-aap/actions/runs/7547915044/job/20548883411?pr=3#step:4:31). The failure happens in the job_resource_test.go

This seems to me like a bad linting error. It's claiming that lines 450-480 of `job_resource_test.go` are duplicates of lines 515-545. They are not, as one is testing idempotence without triggers and the other is testing idempotence with triggers. They are very similar, but clearly not duplicates.

IMO, running anything more than just `go fmt` on tests is not worth it, since tests often are written in ways that code analyzers are likely to flag. I would look at the linters we have running and see if there's a configuration you can use that gets it to ignore this. I'm not sure why this wasn't flagged in the job resource PR.

gravesm (approved):

GomathiselviS:
> > Hi @abikouo Can you help me fix this [lint failure](https://github.com/ansible/terraform-provider-aap/actions/runs/7547915044/job/20548883411?pr=3#step:4:31). The failure happens in the job_resource_test.go
> 
> This seems to me like a bad linting error. It's claiming that lines 450-480 of `job_resource_test.go` are duplicates of lines 515-545. They are not, as one is testing idempotence without triggers and the other is testing idempotence with triggers. They are very similar, but clearly not duplicates.
> 
> IMO, running anything more than just `go fmt` on tests is not worth it, since tests often are written in ways that code analyzers are likely to flag. I would look at the linters we have running and see if there's a configuration you can use that gets it to ignore this. I'm not sure why this wasn't flagged in the job resource PR.

Based on the information provided in https://github.com/mibk/dupl?tab=readme-ov-file#dupl, it appears that the golangci-lint will generate duplicate code (dupl errors) for the test cases in job_resource_test.go

internal/provider/group_resource.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	d.Name = types.StringValue(result["name"].(string))
+	if result["description"] != "" {
+		d.Description = types.StringValue(result["description"].(string))
+	} else {
+		d.Description = types.StringNull()
+	}
+	d.URL = types.StringValue(result["url"].(string))
+	if result["variables"] != nil {
```
hakbailey, line 140: I don't think the variables field in the API response will ever be nil, if there are no variables supplied when the group resource is created then the variables field returned will be an empty string so that's probably what we should check for here.

internal/provider/group_resource.go
```diff
@@ -0,0 +1,345 @@
...
+
+	json_raw, err := json.Marshal(body)
+	if err != nil {
+		diags.Append(diag.NewErrorDiagnostic("Body JSON Marshal Error", err.Error()))
+		return nil, diags
+	}
+	return json_raw, diags
+}
+
+func (d *GroupResourceModel) ParseHttpResponse(body []byte) error {
```
hakbailey, line 124: I think this method should add all of the fields from the response body to the GroupResourceModel, including the inventory id. That way if for some reason the inventory ID in AAP doesn't match what Terraform is expecting, Terraform will error with the appropriate message.

internal/provider/group_resource.go
```diff
@@ -0,0 +1,364 @@
...
+	if IsValueProvided(d.Variables) {
+		// var vars map[string]interface{}
+		// diags.Append(d.Variables.Unmarshal(&vars)...)
+		body["variables"] = d.Variables.ValueString()
+	}
+
+	// URL
+	if IsValueProvided(d.URL) {
+		body["url"] = d.URL.ValueString()
+	}
```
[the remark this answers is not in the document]
hakbailey: Shouldn't the URL *not* be part of the request body? We don't want to send the URL to AAP, that's a field that AAP should manage and return to us.

abikouo (approved):

hakbailey (approved):

PR #4, opened 2024-01-09
Add inventory resource

hakbailey:
Adds an AAP inventory resource. Because there are several open resource PRs, this copies the changed provider and client functionality from those but not the additional changes; I will rebase once those are merged. The changes to inventory_data_source are not related to this new inventory resource, they are due to the copied changes in the client functionality and can be ignored.

To run unit tests, run `go test ./...`. The example file in `examples/resources/inventory` can be used to test this against a running AAP instance. 

Note: I did add some additional functions to the client code (Create, Get, Update, and Delete) as we were duplicating API response checks in every resource. It's still somewhat duplicative so feedback is welcome, but at least now it's only duplicated in one file instead of a bunch.

This PR also deliberately uses a somewhat different code structure and simplified unit tests from the other resource PRs as a starting point for discussion. Again, feedback welcome :)


internal/provider/inventory_resource.go
```diff
@@ -0,0 +1,319 @@
...
+	Variables    jsontypes.Normalized `tfsdk:"variables"`
+}
+
+// generateRequestBody creates a JSON encoded request body from the inventory resource data.
+func (r *inventoryResourceModel) generateRequestBody() ([]byte, diag.Diagnostics) {
+	// Convert inventory resource data to API data model
+	var organizationId int64
+
+	// Use default organization if not provided
+	if r.Organization.ValueInt64() == 0 {
```
abikouo, line 252: not really sure if TF set the default value to `0` while not explicitly provided by the user.
I would rather suggest use both [IsNull()](https://github.com/hashicorp/terraform-plugin-framework/blob/79accaca46cc3989a3defffdc14e8ad3de8f111d/types/basetypes/int64_value.go#L132) and [IsUnknown()](https://github.com/hashicorp/terraform-plugin-framework/blob/79accaca46cc3989a3defffdc14e8ad3de8f111d/types/basetypes/int64_value.go#L137)

internal/provider/inventory_resource.go
```diff
@@ -0,0 +1,319 @@
...
+		organizationId = 1
+	} else {
+		organizationId = r.Organization.ValueInt64()
+	}
+	inventory := inventoryAPIModel{
+		Organization: organizationId,
+		Name:         r.Name.ValueString(),
+		Description:  r.Description.ValueString(),
+		Variables:    r.Variables.ValueString(),
```
abikouo, line 261: these parameters are optional, therefore we should add them to the request only if they are explicitly set in the configuration (based on one test case from your unit tests it seems that it is working like this)

abikouo (commented):
Looks good overall!!

internal/provider/inventory_resource.go
gravesm: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.5.0/types/basetypes#Int64Value.ValueInt64

> If Int64 is null or unknown, returns 0.

internal/provider/inventory_resource.go
hakbailey: Yes, this is because I added `omitempty` to the json struct tags for the `inventoryAPIModel` type. This is Go's built-in way of removing empty fields when JSON-encoding structs.

internal/provider/utils.go
```diff
@@ -1,9 +1,45 @@
...
+	if resp == nil {
+		diags.AddError("HTTP response error", "No HTTP response from server")
+		return diags
+	}
+	if !slices.Contains(expected_statuses, resp.StatusCode) {
+		var info map[string]interface{}
+		err := json.Unmarshal(body, &info)
+		if err != nil {
+			diags.AddError("Error unmarshaling JSON response body", err.Error())
+		}
```
abikouo, line 36: You are trying to unmarshal the body which is not expected to be used anywhere.
Is this part of the ValidateResponse flow? in this case, it should be put outside of this if statement

abikouo (commented):
Just a minor remark

internal/provider/utils.go
```diff
@@ -1,9 +1,45 @@
...
+
 	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 )
 
 func IsValueProvided(value attr.Value) bool {
 	return !value.IsNull() && !value.IsUnknown()
 }
+
+func ValidateResponse(resp *http.Response, body []byte, err error, expected_statuses []int) diag.Diagnostics {
```
gravesm, line 17: I like this function overall, but we are unmarshalling the response twice. I feel like if we renamed it to `ParseResponse()` or something else more appropriate, it could return the marshalled response.

internal/provider/utils.go
hakbailey: The response body is included in the diagnostic detail to give more info on why the unexpected status code was received. But I like Mike's suggestion and will update this function to return the body so we're only unmarshalling it once.

internal/provider/utils.go
hakbailey: Oh, I responded to @abikouo's comment above and then realized why I did not do this. It only gets unmarshalled here if the HTTP status code is not what we expect, so that we can provide the details from the body in the diagnostics error to explain why the unexpected status code was returned. In the inventory_resource `parseHTTPResponse` function it gets unmarshalled directly into the `inventoryApiModel` struct after we know we've received the expected response. So it's only ever getting unmarshalled once, either here _if_ there's an unexpected status code, or in the `parseHTTPResponse` function for each resource (where ultimately I'm hoping we will refactor to unmarshal into a known struct for all the resources).

internal/provider/utils.go
hakbailey: Scratch that, see comment below. I think we can probably ignore this particular error though. If we can't unmarshal the response body here then the diagnostics error simply won't include the body details.

internal/provider/utils.go
gravesm: OK, that makes more sense. Do we know what the response body looks like when the status code isn't one of the expected status codes? Does it ever have anything useful in it in these cases?

hakbailey: Yes, it has useful info at least some of the time. It can vary but provides details like "this resource already exists" or the reason for a bad request, or why there's an authentication error. I added this largely because it was helpful to me for debugging during development and I expect it will be similarly helpful for troubleshooting errors down the line.

abikouo (approved):

alinabuzachis (approved):

gravesm (approved):

PR #7, opened 2024-02-08
Align inventory data source with inventory resource

GomathiselviS:
Refer: https://issues.redhat.com/browse/ACA-1218

With the AAP provider nearing completion, it's essential to realign the inventory data source with the inventory resource. The assignment of hosts and group membership in an inventory is now handled through those resources, making it unnecessary for the inventory data source to represent these. This PR proposes a refactor of the inventory data source to mirror the information available in the inventory resource.

Example configuration file :
```
terraform {
  required_providers {
    aap = {
      source = "ansible/aap"
    }
  }
}

provider "aap" {
  host     = "https://localhost:8043"
  username = "ansible"
  password = "test123!"
  insecure_skip_verify = true
}

resource "aap_inventory" "my_inventory" {
  name = "My new inventory"
  description = "A new inventory for testing"
  variables = jsonencode(
    {
      "foo": "bar"
    }
  )
}

output "inventory" {
  value = aap_inventory.my_inventory
}

data "aap_inventory" "sample" {
  id = aap_inventory.my_inventory.id
}

output "inventory_details" {
  value = data.aap_inventory.sample
}
```

Output of terraform apply:

```
Outputs:

inventory = {
  "description" = "A new inventory for testing"
  "id" = 13
  "name" = "My new inventory"
  "organization" = 1
  "url" = "/api/v2/inventories/13/"
  "variables" = "{\"foo\":\"bar\"}"
}
inventory_details = {
  "description" = "A new inventory for testing"
  "id" = 13
  "name" = "My new inventory"
  "organization" = 1
  "url" = "/api/v2/inventories/13/"
  "variables" = "{\"foo\":\"bar\"}"
}
```


examples/data_source/main.tf
```diff
@@ -13,20 +13,27 @@ provider "aap" {
...
-  type = number
-  description = "The id of the state to request"
+resource "aap_inventory" "my_inventory" {
+  name = "My new inventory"
+  description = "A new inventory for testing"
+  variables = jsonencode(
+    {
+      "foo": "bar"
+    }
+  )
```
abikouo, line 23:
```suggestion
variable "inventory_id" {
   type = number
   description = "The inventory id"
}
```

It is a best practice to keep only the `aap_inventory` data source in this example file as this is one being explained here

examples/data_source/main.tf
```diff
@@ -13,20 +13,27 @@ provider "aap" {
...
-data "aap_inventory" "sample" {
-  path = var.state_id
+output "inventory" {
+  value = aap_inventory.my_inventory
 }
 
-output "inventory_hosts" {
-  value = data.aap_inventory.sample.hosts
+data "aap_inventory" "sample" {
+  id = aap_inventory.my_inventory.id
```
abikouo, line 31:
```suggestion
  id = var.inventory_id
```
combined with the suggestion above

internal/provider/inventory_data_source.go
```diff
@@ -5,102 +5,131 @@ import (
...
+	Id           types.Int64          `tfsdk:"id"`
+	Organization types.Int64          `tfsdk:"organization"`
+	Url          types.String         `tfsdk:"url"`
+	Name         types.String         `tfsdk:"name"`
+	Description  types.String         `tfsdk:"description"`
+	Variables    jsontypes.Normalized `tfsdk:"variables"`
+}
+
+func (d *InventoryDataSource) ReadInventory(id string) (*InventoryDataSourceModel, error) {
+	resp, body, err := d.client.doRequest("GET", "api/v2/inventories/"+id+"/", nil)
```
abikouo, line 76: Please take this opportunity to refactor all code-performing API requests to be aligned with what has been done on the other resources. Here is an example https://github.com/ansible/terraform-provider-aap/blob/d53ac596953fd48d26aa087375b0fbf5c04a6f8b/internal/provider/host_resource.go#L218-L229

abikouo (changes requested):

internal/provider/inventory_data_source.go
```diff
@@ -4,103 +4,143 @@ import (
...
+		return
+	}
 
-	hosts, err := d.ReadAnsibleHosts(state.ID.String())
-	if err != nil {
-		resp.Diagnostics.AddError(
-			"Unable to Read Ansible hosts",
-			err.Error(),
-		)
+	url, diags := getURL(state.Url.ValueString(), state.Id.String())
```
hakbailey, line 116: We can't use `state.Url` here because it won't have a value at this point. Instead the path we provide to `d.client.Get()` should be `"/api/v2/inventories/" + state.Id.String()`.

hakbailey (changes requested):

abikouo (approved):

internal/provider/inventory_data_source.go
```diff
@@ -4,103 +4,97 @@ import (
...
+			"organization": schema.Int64Attribute{
+				Computed:    true,
+				Description: "Identifier for the organization to which the inventory belongs",
+			},
+			"url": schema.StringAttribute{
+				Computed: true,
+			},
+			"name": schema.StringAttribute{
 				Computed: true,
+				Optional: true,
```
hakbailey, line 52: I'm just now realizing that all of the attributes except `id` should be computed and _not_ optional, because the user should never be supplying data for those values in the terraform config. If there is no value for the attribute in AAP, the data source instance will just have a null value for that attribute.

hakbailey (commented):
I think we should probably go ahead and add acceptance tests for the inventory data source.

internal/provider/inventory_data_source_test.go
```diff
@@ -1,269 +1,204 @@
...
+	var inventory InventoryAPIModel
+	randomName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	updatedName := "Inventory " + randomName
+	updatedDescription := "Test inventory"
+	updatedVariables := "{\"foo\": \"bar\"}"
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			// Create and Read testing
```
hakbailey, line 110: For the data source acceptance tests we don't need to test create or update since data sources don't create anything in AAP. Instead, I would suggest creating an inventory in step 1 (using the inventory resource config) and then doing a read test in step 2 with a data source config, using the id from the inventory created in step 1. In step 2 we don't need to check that the resource exists, just that the data source attribute values match what we expect. Let me know if you have questions about any of that.

GomathiselviS: From my understanding, I have addressed your comments. Please let me know if this aligns with your expectations.

hakbailey (commented):
Just a few more things to get the updated acceptance tests passing. Thanks!

PR #11, opened 2024-02-16
Generate docs for terraform-provider-aap

GomathiselviS:
Refer: https://issues.redhat.com/browse/ACA-1230

This PR adds documents for the provider using https://github.com/hashicorp/terraform-plugin-docs


docs/resources/group.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `inventory_id` (Number)
```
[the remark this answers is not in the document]
hakbailey: We should probably also add a basic tools.go file as per the tfplugindocs documentation so that the package is included in go.mod.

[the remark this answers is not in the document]
GomathiselviS: > I would suggest we go ahead and add descriptions of all these attributes in the resource/data source schemas. Generating the docs makes it evident that they are currently a little sparse.

For the inventory data source, I have added the description in https://github.com/ansible/terraform-provider-aap/pull/7

examples/provider/provider.tf
```diff
@@ -0,0 +1,48 @@
...
+  name         = "tf_host"
+  variables = jsonencode(
+    {
+      "foo" : "bar"
+    }
+  )
+  groups = [2, 3, 4]
+}
+
+resource "aap_group" "my_group" {
```
alinabuzachis, line 40: I guess if would be nice to have the example like:
- Create group
- Pass the id of the created group to the host into groups.

alinabuzachis (approved):

hakbailey:
I don't think the Makefile has been updated with the new command? I'm also wondering if we should go ahead and add a Github Actions workflow to generate the docs? That we we can be sure they are always updated when code is changed.

GomathiselviS:
Generate-docs action is verified [here](https://github.com/ansible/terraform-provider-aap/actions/runs/7993392787)

GomathiselviS:
> I don't think the Makefile has been updated with the new command? I'm also wondering if we should go ahead and add a Github Actions workflow to generate the docs? That we we can be sure they are always updated when code is changed.

I have added the commands to Makefile and added a GH workflow.

.github/workflows/generate_docs.yml
```diff
@@ -0,0 +1,29 @@
+name: Generate Terraform Plugin Documentation
```
hakbailey, line 1: Correct me if I'm wrong but I think this is ephemeral? It will generate the docs but they aren't committed to the repo so any changes won't get merged. I'm thinking what we probably want here is to have this workflow compare the generated docs to what is in the repo to ensure that any updated docs have been committed?

.github/workflows/generate_docs.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: go get github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs@latest
+
+      - name: Build
+        run: go build -v ./...
+
+      - name: Generate documentation
+        run: make generatedocs
+
+      - name: Use GITHUB_TOKEN to commit changes
+        run: |
```
alinabuzachis, line 32: I would probably keep an approach where commits to docs are only made in case of prep release PR.  So the documents are automatically updated only in case of release. I this way we limit number of commits to those files and unwanted manual tweaks that may occur. In other cases, we can only render docs and ensure that the generation process works without any issue. If issues occur, just give a log to see the root cause of the issue. But, this is only my thought on this process.

.github/workflows/generate_docs.yml
GomathiselviS: I have added a check to see if there are any diffs.

hakbailey (approved):

.github/workflows/generate_docs.yml
GomathiselviS: Considering our discussions, I am proceeding to merge this PR. Should you feel the need for alterations in this workflow, kindly create an issue, and we can work on the fix.

PR #12, opened 2024-02-21
Ensure variables and extra_vars are always treated as strings

alinabuzachis:
Ensure variables and extra_vars are always treated as strings


abikouo (approved):

hakbailey (commented):
I would probably add to the `Description` for each of these changed attributes that the variables must be provided as either a JSON or YAML string. We might also want to update the examples to show a YAML example for variables so users know what that should look like. 

This might be overkill but do we want to include YAML variables in our acceptance tests as well as JSON?

abikouo:
I dont know if it is related to this pull request, but I got the following issue

```terraform
variable "inventory_name" {
  type = string
  default = "aca-1272-2"
}

resource "aap_inventory" "main" {
  organization = 1
  name = var.inventory_name
  description = "An inventory created using Terraform provider for AAP"
  variables = yamlencode({"os": "Linux", "automation": "ansible"})
}
```

running terraform apply I got the following error

```error
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to aap_inventory.main, provider "provider[\"registry.terraform.io/ansible/aap\"]" produced an unexpected new value:
│ .variables: was cty.StringVal("\"automation\": \"ansible\"\n\"os\": \"Linux\"\n"), but now cty.StringVal("\"automation\": \"ansible\"\n\"os\":
│ \"Linux\"").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

abikouo (commented):
It would be good to update examples to show how to create resources defining variables as YAML, something like `variables =yamlencode({"os"="RedHat", "automation"="ansible-milestone"})`

abikouo (changes requested):
It would be good to update examples to show how to create resources defining variables as YAML, something like `variables =yamlencode({"os"="RedHat", "automation"="ansible-milestone"})`

gravesm:
> The inventory has been created into AAP as expected, however the terraform execution reports a failure

The problem is that the string that you set in your config has to *exactly* match the end state that is stored in the statefile for a default string type attribute. It seems like AAP is stripping the newline from the end of the string, and this is what we're storing in the statefile. I'm struggling to think of a good solution for this problem. If we're going to store this field as a string, we will need to mimic whatever string processing AAP is doing on this.

We could either strip the newline from what's provided in the config, or add a newline to what we get from AAP. Adding a newline feels weird to me, and could potentially lead to other problems. Based on https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md, I don't think TF will let us alter the config. Unfortunately, I think the only solution is to create a custom type here, and add logic to not read trailing newlines as a change that would trigger an update.

I'm open to other suggestions, though.

abikouo:
> > The inventory has been created into AAP as expected, however the terraform execution reports a failure
> 
> The problem is that the string that you set in your config has to _exactly_ match the end state that is stored in the statefile for a default string type attribute. It seems like AAP is stripping the newline from the end of the string, and this is what we're storing in the statefile. I'm struggling to think of a good solution for this problem. If we're going to store this field as a string, we will need to mimic whatever string processing AAP is doing on this.
> 
> We could either strip the newline from what's provided in the config, or add a newline to what we get from AAP. Adding a newline feels weird to me, and could potentially lead to other problems. Based on https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md, I don't think TF will let us alter the config. Unfortunately, I think the only solution is to create a custom type here, and add logic to not read trailing newlines as a change that would trigger an update.
> 
> I'm open to other suggestions, though.

from what I am understanding is that Terraform is complaining because this attribute is defined as `input` (not computed) but here we are trying to compute it https://github.com/alinabuzachis/terraform-provider-aap/blob/76e51cd7935e754a23a16c7658042718661b3350/internal/provider/inventory_resource.go#L299
This means, the resource state going to be saved differs from what Terraform expected. Here we are talking about a trailing line but any update AAP will perform on this attribute will lead to this error. 
The easiest way for me to fix it, is to remove this update from data received from AAP (This is valid for all others attributes considered as inputs)

alinabuzachis:
> > > The inventory has been created into AAP as expected, however the terraform execution reports a failure
> > 
> > 
> > The problem is that the string that you set in your config has to _exactly_ match the end state that is stored in the statefile for a default string type attribute. It seems like AAP is stripping the newline from the end of the string, and this is what we're storing in the statefile. I'm struggling to think of a good solution for this problem. If we're going to store this field as a string, we will need to mimic whatever string processing AAP is doing on this.
> > We could either strip the newline from what's provided in the config, or add a newline to what we get from AAP. Adding a newline feels weird to me, and could potentially lead to other problems. Based on https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md, I don't think TF will let us alter the config. Unfortunately, I think the only solution is to create a custom type here, and add logic to not read trailing newlines as a change that would trigger an update.
> > I'm open to other suggestions, though.
> 
> from what I am understanding is that Terraform is complaining because this attribute is defined as `input` (not computed) but here we are trying to compute it https://github.com/alinabuzachis/terraform-provider-aap/blob/76e51cd7935e754a23a16c7658042718661b3350/internal/provider/inventory_resource.go#L299 This means, the resource state going to be saved differs from what Terraform expected. Here we are talking about a trailing line but any update AAP will perform on this attribute will lead to this error. The easiest way for me to fix it, is to remove this update from data received from AAP (This is valid for all others attributes considered as inputs)

From my understanding, `Computed: true` is used for all those values that are not user configurable or known after apply. However, this should not be the problem because variables are not marked as Computed. The problem is a discrepancy between what TF expects and what the API returns in the case of YAML strings (as you can see in the log you pasted). The returned value for the variables field does not contained the new line char at the end.

I did not understand what you mean by `Here we are talking about a trailing line but any update AAP will perform on this attribute will lead to this error.`. Can you please clarify?

However, I see that removing the value interpolation here https://github.com/alinabuzachis/terraform-provider-aap/blob/76e51cd7935e754a23a16c7658042718661b3350/internal/provider/inventory_resource.go#L299 might solve the issue.

gravesm:
> The easiest way for me to fix it, is to remove this update from data received from AAP (This is valid for all others attributes considered as inputs)

This isn't going to work. You are effectively saying that what's in the config is the state of the resource. This won't be the case if someone manually changes the variables from the AAP UI. If this happens, a `tf plan` will show no changes needed even though the desired state and the actual state are different. AAP should always be the authority for what the actual state of the resource is.

abikouo:
> > The easiest way for me to fix it, is to remove this update from data received from AAP (This is valid for all others attributes considered as inputs)
> 
> This isn't going to work. You are effectively saying that what's in the config is the state of the resource. This won't be the case if someone manually changes the variables from the AAP UI. If this happens, a `tf plan` will show no changes needed even though the desired state and the actual state are different. AAP should always be the authority for what the actual state of the resource is.

You are right, I missed this use case when the user is updating directly on AAP

abikouo:
> I did not understand what you mean by `Here we are talking about a trailing line but any update AAP will perform on this attribute will lead to this error.`. Can you please clarify?
>
@alinabuzachis What I am trying to explain is that here we are seeing that AAP is removing the trailing line but I don't know if it is the only update it is doing on the content of the variables, otherwise, we need to consider data manipulation done on AAP

internal/provider/customtypes/customstring_type.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+var (
+	_ = basetypes.StringTypable(&CustomStringType{})
+	_ = xattr.TypeWithValidate(&CustomStringType{})
+)
+
+// CustomStringType implements a custom Terraform type.
+type CustomStringType struct {
```
abikouo, line 21: What if you add an AAP prefix `AAPCustomStringType`, to specify that this is not implemented on Terraform provider plugin?

abikouo (approved):

internal/provider/customtypes/customstring_type.go
alinabuzachis: Renamed using the AAP prefix.

examples/resources/group/main.tf
```diff
@@ -7,18 +7,70 @@ terraform {
...
+  inventory_id = 1
+  name         = "tf_group_abc"
+  variables    = yamlencode({ "os" : "Linux", "automation" : "ansible" })
+}
+
+resource "aap_group" "sample_xyz" {
+  inventory_id = 1
+  name         = "tf_group_xyz"
+  variables    = "Not valid JSON"
+}
```
gravesm, line 56: Why include this one?

internal/provider/inventory_data_source.go
```diff
@@ -59,8 +59,8 @@ func (d *InventoryDataSource) Schema(_ context.Context, _ datasource.SchemaReque
 			},
 			"variables": schema.StringAttribute{
 				Computed:    true,
-				CustomType:  jsontypes.NormalizedType{},
-				Description: "Variables of the inventory",
+				CustomType:  customtypes.AAPCustomStringType{},
+				Description: "Variables of the inventory. Must be provided as either a JSON or YAML string.",
```
gravesm, line 63: Since this is a data source, the user isn't providing this attribute. I would either revert this change or have it say something like "Will be either JSON or YAML string depending on how the variables were entered into AAP."

examples/resources/group/main.tf
alinabuzachis: This is a copy-paste issue. It should be "os: Linux\nautomation: ansible-devel". I'll fix that, Thanks.

internal/provider/inventory_data_source.go
alinabuzachis: I modified the description with your suggestion.

gravesm (approved):

PR #30, opened 2024-06-19
Support dynamic AAP API endpoints

abikouo:
JIRA issue: [ACA-1609](https://issues.redhat.com/browse/ACA-1609)


internal/provider/inventory_data_source.go
```diff
@@ -78,7 +80,9 @@ func (d *InventoryDataSource) Read(ctx context.Context, req datasource.ReadReque
 		return
 	}
 
-	readResponseBody, diags := d.client.Get("api/v2/inventories/" + state.Id.String())
+	resourceURL := path.Join(d.client.getApiEndpoint(), "inventories", state.Id.String())
+	tflog.Info(ctx, fmt.Sprintf("Inventory resource url: (%s)", resourceURL))
```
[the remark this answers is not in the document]
abikouo: I have set them just for relevant information.
I can set them at `debug` level.

internal/provider/provider.go
```diff
@@ -132,16 +132,8 @@ func (p *aapProvider) Configure(ctx context.Context, req provider.ConfigureReque
...
-		resp.Diagnostics.AddError(
-			"Unable to Create AAP API Client",
-			"An unexpected error occurred when creating the AAP API client. "+
-				"If the error is not clear, please contact the provider developers.\n\n"+
-				"http Client Error: "+err.Error(),
-		)
-		return
-	}
+	client, diags := NewClient(host, &username, &password, insecureSkipVerify, timeout)
+	resp.Diagnostics.Append(diags...)
```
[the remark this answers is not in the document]
abikouo: we have reach the end of the function, whether the `resp.Diagnostics` has error or not, we will exit the same way so no need to add this check here

internal/provider/client.go
```diff
@@ -42,14 +45,38 @@ func NewClient(host string, username *string, password *string, insecureSkipVeri
 	}
 	client.httpClient = &http.Client{Transport: tr, Timeout: time.Duration(timeout) * time.Second}
 
-	return &client, nil
+	body, diags := client.Get("/api/")
```
gravesm, line 48: I think what we really need to do is support accessing controller directly (AAP 2.4) and accessing controller through the API gateway (AAP 2.5+). What we have here works for controller directly, but not for gateway. We'll probably need to assume gateway, first, then if that fails, fallback to controller directly. Here's what I get when I access `/api/` on gateway:
```json
{
  "description": "AAP gateway REST API",
  "apis": {
    "gateway": "/api/gateway/",
    "controller": "/api/controller/",
    "eda": "/api/eda/",
    "galaxy": "/api/galaxy/"
  }
}
```
A further fetch of `/api/controller/` gives:
```json
{
  "description": "AWX REST API",
  "current_version": "/api/controller/v2/",
  "available_versions": {
    "v2": "/api/controller/v2/"
  },
  "custom_logo": "",
  "custom_login_info": "",
  "login_redirect_override": ""
}
```

abikouo: updated

internal/provider/client_test.go
```diff
@@ -22,12 +25,122 @@ func TestComputeURLPath(t *testing.T) {
...
+	if len(expectedErr) > 0 {
+		if err == nil || expectedErr != err.Error() {
+			t.Fatalf("Expected error (%s) - current error (%v)", expectedErr, err)
+		}
+	} else {
+		if err != nil {
+			t.Fatalf("Unexpected error (%v)", err)
+		}
+		assert.Equal(t, result, expectedValue, fmt.Sprintf("Expected value (%v), got (%v)", expectedValue, result))
+	}
```
gravesm, line 51:
```suggestion
	if len(expectedErr) > 0 {
		if assert.Error(t, err) {
		        assert.Equal(t, expectedErr, err)
		}
	} else {
		if assert.NoError(t, err) {
		        assert.Equal(t, result, expectedValue, fmt.Sprintf("Expected value (%v), got (%v)", expectedValue, result))
		}
	}
```

alinabuzachis (approved):

gravesm (approved):

internal/provider/client.go
```diff
@@ -18,18 +20,72 @@ type ProviderHTTPClient interface {
 	Get(path string) ([]byte, diag.Diagnostics)
 	Update(path string, data io.Reader) ([]byte, diag.Diagnostics)
 	Delete(path string) ([]byte, diag.Diagnostics)
+	setApiEndpoint() diag.Diagnostics
+	getApiEndpoint() string
+}
+
+func GetKeyFromJson[T interface{}](jsonData []byte, keyName string, value *T) error {
```
hakbailey, line 27: This seems like more of a utility function than a client function, would it makes sense to move it to utils.go?

internal/provider/client_test.go
```diff
@@ -22,12 +25,121 @@ func TestComputeURLPath(t *testing.T) {
...
+			case int64:
+				var res int64
+				err := GetKeyFromJson[int64](jsonData, tc.keyName, &res)
+				assertTest[int64](t, tc.err, err, v, res)
+			}
+		})
+	}
+}
+
+type mEndPointHttpClient struct {
```
hakbailey, line 81: Particularly for testing our client code, I dislike the approach of mocking all the client functions and would prefer that we use the built-in httptest package as describe [here](https://medium.com/zus-health/mocking-outbound-http-requests-in-go-youre-probably-doing-it-wrong-60373a38d2aa). What do you think?

internal/provider/client_test.go
```diff
@@ -22,12 +25,121 @@ func TestComputeURLPath(t *testing.T) {
...
+	}{
+		{name: "AAP 2.4", mEndpoint: map[string][]byte{"/api/": endPoint24}, expected: "/api/v2/"},
+		{name: "AAP 2.5+", mEndpoint: map[string][]byte{"/api/": gateway25, "/api/controller/": controller25}, expected: "/api/controller/v2/"},
+	}
+	for _, tc := range testTable {
+		t.Run(tc.name, func(t *testing.T) {
+			client := &mEndPointHttpClient{Endpoints: tc.mEndpoint}
+			result, diags := readApiEndpoint(client)
+			assert.Equal(t, diags.HasError(), false, fmt.Sprintf("readApiEndpoint() returns errors (%v)", diags))
+			assert.Equal(t, result, tc.expected, fmt.Sprintf("Returned value differ - expected value (%s), got (%v)", tc.expected, result))
```
hakbailey, line 142: I think in assert.Equal the params are supposed to be supplied (t, expected, actual, ...) otherwise the built-in error message displays the values reversed.

abikouo: updated.

internal/provider/client_test.go
abikouo: updated

hakbailey (approved):

PR #32, opened 2024-06-25
Fix inventoryID issue with Job resource

### LINKED ISSUE #31: Failed to update job resource with inventory Id
Steps to reproduce:

main.tf
```tf
variable "template_id" {
  type        = number
}

resource "aap_job" "testing" {
  job_template_id   = var.template_id
}
```
Apply the above configuration `terraform apply --auto-approve`
update the main.tf configuration using 

```tf
variable "template_id" {
  type        = number
}

resource "aap_inventory" "testing" {
  name = "my_inventory"
}

resource "aap_job" "testing" {
  job_template_id   = var.template_id
  inventory_id = aap_inventory.testing.id
}
```
Apply the above configuration `terraform apply --auto-approve`, the following issue appears

```shell
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to aap_job.testing, provider "provider[\"registry.terraform.io/ansible/aap\"]" produced an unexpected new value:
│ .inventory_id: was cty.NumberIntVal(xx), but now cty.NumberIntVal(1).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

**_Notes_**: This is reproducible only if the job template allows specifying the inventory ID at launch.

abikouo:
Closes #31 
- Fix issue when create Job resource using Job template ID which does not allow running with a different inventory ID (Template does not prompt for inventory Id at launch)
- Update acceptance tests to validate inventory Id


internal/provider/job_resource.go
```diff
@@ -267,7 +267,10 @@ func (r *JobResourceModel) ParseHttpResponse(body []byte) diag.Diagnostics {
 	r.URL = types.StringValue(resultApiJob.URL)
 	r.Status = types.StringValue(resultApiJob.Status)
 	r.TemplateID = types.Int64Value(resultApiJob.TemplateID)
-	r.InventoryID = types.Int64Value(resultApiJob.Inventory)
+	if r.InventoryID.ValueInt64() == 0 {
+		r.InventoryID = types.Int64Value(resultApiJob.Inventory)
+	}
```
gravesm, line 272: I feel like the bug this is trying to fix was introduced by https://github.com/ansible/terraform-provider-aap/pull/10/files#diff-afd9b829236927b85ef1814941bb6f012eed9cca71547f9f9ce15dcceb29a6f4R216-R219. I don't know why that change was made, but, among other things, it makes it impossible to use a job template that has a default inventory and is configured not to prompt on launch. I think fixing that bug makes this bug go away, because it shouldn't be possible to launch a job with prompt on launch and no default inventory set.

Issue #36, opened 2024-08-12
Set a Default Organization ID at the Provider Level


benemon:
It would be beneficial to provider the option to specify a Default Organization at the Provider level within the Provider.

In much the same way that the Provider currently defaults to the RHAAP Default Organization value of `1` where no Organization attribute is supplied, it would useful for Organizational automation to be able to default all Organization-aware resources to something other than `1`.

In addition, in much the same way that setting an Organization at the Resource level overrides the implicit Organization today, and explicit Resource attribute being set for Organization should also be honoured.

This becomes more relevant as the different types of RHAAP Resource become supported by the Provider.

PR #37, opened 2024-08-12
Initial pass of default_organization support

benemon:
Initial attempt at addressing #36.


benemon:
Whilst undertaking this work, I have noticed an inconsistency in how the Provider manages deltas between RHAAP Resources and local state. I have verified this with a "vanilla" build of the current Provider version. I will raise a separate issue on that.

dleehr:
I'm going to go ahead and close this since #107 implements a data source that addresses this gap.

PR #49, opened 2025-01-14
Release prep for v1.1.0

mandar242:
Release prep for v1.1.0


gravesm:
I think we should update the go version in go.mod to 1.23.0, and then do a dependency update. We need to make sure that all the integration tests are passing locally. Probably worth making sure the versions in CI workflows are updated, too, though I think as long as it's 1.21 it shouldn't really matter. We should do this in another PR.

mandar242:
> go version in go.mod to 1.23.0, and then do a dependency update.

@gravesm would go version update be a blocker for the v1.1.0 release?

mandar242:
> I think we should update the go version in go.mod to 1.23.0, and then do a dependency update. We need to make sure that all the integration tests are passing locally. Probably worth making sure the versions in CI workflows are updated, too, though I think as long as it's 1.21 it shouldn't really matter. We should do this in another PR.

https://github.com/ansible/terraform-provider-aap/pull/50
@gravesm

hakbailey (approved):

changelogs/changelog.yaml
```diff
@@ -8,3 +8,11 @@ releases:
...
     - 1.0.0.yml
     release_date: '2024-05-01'
+  1.1.0:
+    changes:
+      minor_changes:
+      - Support dynamic value for AAP endpoints since the value depends on the AAP
+        version (https://github.com/ansible/terraform-provider-aap/pull/30).
+    fragments:
+    - 20240619-aap-versioning.yaml
+    release_date: '2025-01-14'
```
alinabuzachis, line 18: Can you please adjust the release date?

alinabuzachis (approved):

gravesm (approved):

PR #50, opened 2025-01-14
update to go 1.23.0, update dependencies

mandar242:
- Update Go version to 1.23.0
- Update Go version in workflows to 1.23.0
- Update linter workflow to use latest stable `actions/checkout@v4`, `actions/setup-go@v5`, `golangci/golangci-lint-action@v6`
- linter fixes to pass ci linter, including removal of deprecated interface `xattr.TypeWithValidate()`

Imo, we can safely remove `xattr.TypeWithValidate` as [func (t AAPCustomStringType) Validate(_ context.Context, in tftypes.Value, path path.Path) diag.Diagnostics {](https://github.com/ansible/terraform-provider-aap/blob/main/internal/provider/customtypes/aapcustomstring_type.go#L74-L111) implements the type validation


internal/provider/customtypes/aapcustomstring_type.go
```diff
@@ -14,7 +13,6 @@ import (
 
 var (
 	_ = basetypes.StringTypable(&AAPCustomStringType{})
-	_ = xattr.TypeWithValidate(&AAPCustomStringType{})
```
gravesm, line 17: This may get the linters to pass, but if this is deprecated, presumably the `Validate()` method will stop being called by TF at some point.

mandar242: what would be the preferred approach for this?
1. Ignore lint failure on the deprecated (only failure in this PR is xattr.TypeWithValidate being deprecated)
2. Replace deprecated interface with newer [xattr.ValidateableAttribute interface](https://developer.hashicorp.com/terraform/plugin/framework/validation#defining-type-validation:~:text=information%20about%20using-,xattr.ValidateableAttribute%20interface,-%2C%20and%20function.ValidateableParameter) or [function.ValidateableParameter interface instead.](https://developer.hashicorp.com/terraform/plugin/framework/validation#defining-type-validation:~:text=interface%2C%20and-,function.ValidateableParameter%20interface,-instead.)

If we need to implement newer implement I would appreciate guidance on that, as we have to get this merged to proceed with release as soon as we can.
@gravesm @hakbailey

gravesm (approved):
I'll go ahead and approve this to not block releasing, but it's really important that we deal with the deprecation. By removing the type check on the interface we've removed the only warning we're going to get about this.

internal/provider/customtypes/aapcustomstring_type.go
hakbailey: Following on Mike's earlier comment, I vote we go ahead with the release as is but create a Jira ticket to deal with the deprecation.

hakbailey (approved):

PR #53, opened 2025-01-20
Update release workflow - .goreleaser.yml and release.yml

mandar242:
Update `.goreleaser.yml` and `.github/workflows/release.yml` to fix release failure.
https://github.com/ansible/terraform-provider-aap/actions/runs/12873507516/job/35891192874

Updated https://github.com/ansible/terraform-provider-aap/blob/main/.goreleaser.yml and https://github.com/ansible/terraform-provider-aap/blob/main/.github/workflows/release.yml

with 
https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.goreleaser.yml and https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.github/workflows/release.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,22 +17,22 @@ jobs:
   goreleaser:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           # Allow goreleaser to access older tag information.
           fetch-depth: 0
-      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+      - uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
         with:
           go-version-file: 'go.mod'
           cache: true
       - name: Import GPG key
-        uses: crazy-max/ghaction-import-gpg@01dd5d3ca463c7f10f7f4f7b4f177225ac661ee4 # v6.1.0
+        uses: crazy-max/ghaction-import-gpg@cb9bde2e2525e640591a934b1fd28eef1dcaf5e5 # v6.2.0
         id: import_gpg
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -17,22 +17,22 @@ jobs:
   goreleaser:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
hakbailey, line 20: Is there a reason we're using hashes here instead of the action versions? Usually we would just do `actions/checkout@v4`, etc.

.goreleaser.yml
```diff
@@ -16,9 +17,18 @@ builds:
...
     - linux
+    - darwin
   goarch:
     - amd64
+    - '386'
+    - arm
+    - arm64
+  ignore:
+    - goos: darwin
+      goarch: '386'
```
gravesm, line 31: These changes should all be reverted as we're only building linux amd64.

.github/workflows/release.yml
gravesm: It's copied/pasted from https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.github/workflows/release.yml is why. Whether we want to or not, I don't really have a strong opinion about.

mandar242: I am not really sure on that.
The current file in the codebase uses hashes https://github.com/ansible/terraform-provider-aap/blob/main/.github/workflows/release.yml
and also one from [terraform-provider-scaffolding-framework](https://github.com/hashicorp/terraform-provider-scaffolding-framework) uses hashes https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.github/workflows/release.yml
so continued with the same

.goreleaser.yml
mandar242: removed.

.github/workflows/release.yml
hakbailey: Pinning the major version will automatically use minor and patch updates without us having to update it, which I am in favor of.

gravesm (approved):

PR #61, opened 2025-05-06
Fixes issue where a deleted AAP Job causes terraform plan to fail

dleehr:
Updates the behavior for the `aap_job` resource to handle a 404 error in the `Read` method. A 404 error indicates the job recorded in terraform state was created by the provider but has since been deleted from AAP. As noted in #59, this scenario is common because AAP cleans up old jobs by default and results in plan failures that require manual state modification.

With this change, we remove the resource from the state if a 404 is encountered, so that the plan marks it as disappeared and plans to re-create. On apply, it is re-created. I've verified the acceptance tests locally against AAP 2.5.

This PR also updates the regex patterns in the acceptance tests to support AAP 2.5 and configured `main.go` to support debugging with a `-debug` flag.

Fixes #59

Issue #52, opened 2025-01-17
Add a resource to sync inventory


amolari:
Please add a resource that will sync a specific inventory with the possibility to either:

1. sync all the sources of the inventory
2. sync specific sources (list) of the inventory

That resource would be helpful if on the Terraform side one provision/deprovision some resources that are part of the inventory.

PR #62, opened 2025-05-06
Inventory Enhancements (Resource and Data Source)

lranjbar:
Part of [AAP-44426](https://issues.redhat.com/browse/AAP-44426)

User Story
As a user of Terraform and AAP, I want the [terraform-provider-aap](https://registry.terraform.io/providers/ansible/aap) to implement [data sources](https://developer.hashicorp.com/terraform/language/data-sources) for the Job Template and Inventory objects in AAP so that I can look up existing objects when creating AAP jobs from Terraform.

### DIFF
--- a/docs/resources/inventory.md
+++ b/docs/resources/inventory.md
@@ -114,3 +114,29 @@ output "inventory_xyz" {
 - `named_url` (String) Named URL of the inventory
 - `organization_name` (String) Name for the organization.
 - `url` (String) URL of the inventory
+
+## Inventory Look Up
+
+You can look up inventories by using either the `id` or a combination of `name` and `organization_name`.
+
+Creating a new inventory in the Default organization:
+```terraform
+resource "aap_inventory" "sample" {
+  name        = "My Sample Inventory"
+  organization_name = "Default"
+  description = "A new inventory for testing"
+  variables   = jsonencode(yamldecode(local.values_variables))
+}
+```
+
+You can access this inventory using either the `id` or the combination of `name` and `organization_name`.
+```terraform
+data "aap_inventory" "sample" {
+  id = aap_inventory.sample.id
+}
+
+data "aap_inventory" "sample" {
+  name = aap_inventory.sample.name
+  organization_name = aap_inventory.sample.organization_name
+}
+```
--- a/templates/resources/inventory.md.tmpl
+++ b/templates/resources/inventory.md.tmpl
@@ -1,0 +1,43 @@
+---
+page_title: "{{ .Name }} {{ .Type }} - {{ .ProviderName }}"
+description: |-
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{ .Name }} ({{ .Type }})
+
+{{ .Description | trimspace }}
+
+{{ if .HasExample }}
+## Example Usage
+
+{{ tffile .ExampleFile }}
+{{ end }}
+
+{{ .SchemaMarkdown | trimspace }}
+
+## Inventory Look Up
+
+You can look up inventories by using either the `id` or a combination of `name` and `organization_name`.
+
+Creating a new inventory in the Default organization:
+```terraform
+resource "aap_inventory" "sample" {
+  name        = "My Sample Inventory"
+  organization_name = "Default"
+  description = "A new inventory for testing"
+  variables   = jsonencode(yamldecode(local.values_variables))
+}
+```
+
+You can access this inventory using either the `id` or the combination of `name` and `organization_name`.
+```terraform
+data "aap_inventory" "sample" {
+  id = aap_inventory.sample.id
+}
+
+data "aap_inventory" "sample" {
+  name = aap_inventory.sample.name
+  organization_name = aap_inventory.sample.organization_name
+}
+```


docs/data-sources/inventory.md
```diff
@@ -45,14 +45,16 @@ output "inventory_details" {
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
-### Required
+### Optional
```
AaronH88, line 48: Was this auto generated? 

At least one of these 3 fields is required right?

internal/provider/inventory_data_source.go
```diff
@@ -79,7 +105,14 @@ func (d *InventoryDataSource) Read(ctx context.Context, req datasource.ReadReque
 		return
 	}
 
-	resourceURL := path.Join(d.client.getApiEndpoint(), "inventories", state.Id.String())
+	URI := path.Join(d.client.getApiEndpoint(), "inventories")
+	resourceURL, err := ReturnAAPNamedURL(state.Id, state.Name, state.OrganizationName, URI)
+	// resourceURL, err := state.ResourceUrlFromParameters(d)
```
AaronH88, line 110: This function Read takes in a context but we dont watch it, any reason why? 

Also a commented out bit of code here 😄

internal/provider/inventory_data_source.go
```diff
@@ -118,17 +151,63 @@ func (d *InventoryDataSource) Configure(_ context.Context, req datasource.Config
...
+			datasourcevalidator.AtLeastOneOf(
+				tfpath.MatchRoot("id")),
+			datasourcevalidator.RequiredTogether(
+				tfpath.MatchRoot("name"),
+				tfpath.MatchRoot("organization_name")),
+		),
+	}
+}
+
+func (d *InventoryDataSource) ValidateConfig(ctx context.Context, req provider.ValidateConfigRequest, resp *provider.ValidateConfigResponse) {
```
AaronH88, line 167: Am i right in guessing that these structs can not be replaced by interfaces?

internal/provider/inventory_data_source.go
```diff
@@ -118,17 +151,63 @@ func (d *InventoryDataSource) Configure(_ context.Context, req datasource.Config
...
+
+	if IsValueNotProvided(data.Id) && IsValueNotProvided(data.Name) {
+		resp.Diagnostics.AddAttributeWarning(
+			tfpath.Root("id"),
+			"Missing Atribute Configuration",
+			"Expected either [id] or [name + organization_name] pair",
+		)
+	}
+
+	if IsValueProvided(data.Name) && IsValueNotProvided(data.OrganizationName) {
```
AaronH88, line 192: Which is easier to read, two functions with very similar names but complete opposite returns, or, `!IsValueProvided()`

internal/provider/utils.go
```diff
@@ -15,8 +16,23 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
+func ReturnAAPNamedURL(id types.Int64, name types.String, orgName types.String, URI string) (string, error) {
+	if IsValueProvided(id) {
```
AaronH88, line 20: because each of these cases is immediately returning, there is no need for the `if, else if` blocks 

Also, is there ever a case where we want all 3, `id, name, orgName` in the URI?

AaronH88:
Overall this is looking good, 
I have a common question around checking contexts throughout the provider, is this a pattern in providers, to ignore contexts? Or is this just a small bit of tec debt from the first implementation? 

Also, some of the functions seem like their arguments could be replaced with interfaces, only if, terraform dosent try to use reflection when calling these functions. 😄

internal/provider/inventory_data_source.go
AaronH88: No need for a poll, its not that important, and ok the way it is. 

I personally find the `!` more readable but its grand the way it is

docs/data-sources/inventory.md
```diff
@@ -45,14 +45,16 @@ output "inventory_details" {
...
 
 - `id` (Number) Inventory id
+- `name` (String) Name of the inventory
+- `organization_name` (String) The name for the organization to which the inventory belongs
 
 ### Read-Only
 
 - `description` (String) Description of the inventory
-- `name` (String) Name of the inventory
+- `named_url` (String) The Named Url of the inventory
```
dleehr, line 57: I'm not sure what a `named_url` is and couldn't find it in the API docs. 

Is this a pattern something we're establishing? Can we show in the examples what this would look like?

docs/data-sources/inventory.md
dleehr: Technically that might be the only option - making them all optional to support lookups. But I think we should update the docs (possibly in the template that updates this file) to make it clear that you either need id or name+org

dleehr (commented):
Really excited to see the rapid progress here and impressed at the turnaround.

I had some questions about the `named_url`, but more importantly I think we need to add acceptance tests for this functionality.

docs/data-sources/inventory.md
tyraziel: Information about `named_url` can be found here:

[2.5 documentation - Automation execution API overview Chapter 8](https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.5/html/automation_execution_api_overview/controller-api-access-resources)

[2.4 documentation - Automation Controller API overview Chapter 8](https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.4/html/automation_controller_api_overview/controller-api-access-resources)

docs/data-sources/inventory.md
tyraziel: Yes, agree, the docs should be updated to reflect this.

Makefile
```diff
@@ -18,6 +18,10 @@ testacc:
 	@echo "==> Running acceptance tests..."
 	TF_ACC=1 AAP_HOST="https://localhost:8043" AAP_INSECURE_SKIP_VERIFY=true go test -count=1 -v ./...
 
+testacc-aapdev:
+	@echo "==> Running acceptance tests..."
+	TF_ACC=1 AAP_HOST="http://localhost:9080" AAP_INSECURE_SKIP_VERIFY=true go test -count=1 -v ./...
```
dleehr, line 23: Nitpick but if we're connecting over `http` and not `https`, the `AAP_INSECURE_SKIP_VERIFY` flag isn't necessary. See https://github.com/ansible/terraform-provider-aap/pull/61#issuecomment-2854833869

```suggestion
	TF_ACC=1 AAP_HOST="http://localhost:9080" go test -count=1 -v ./...
```

docs/data-sources/inventory.md
dleehr: I see this conversation was resolved but there weren't any changes. Was there a resolution?

docs/data-sources/inventory.md
dleehr: Similar question as the previous - I see the conversation was resolved after we discussed linking to the API reference but don't see any changes. What was the resolution?

internal/provider/job_resource_test.go
```diff
@@ -317,6 +317,7 @@ func TestAccAAPJob_UpdateWithSameParameters(t *testing.T) {
 }
 
 func TestAccAAPJob_UpdateWithNewInventoryId(t *testing.T) {
+	t.Skip("Test seems to always create new job in the default inventory after adding inventory data source.")
```
dleehr, line 320: This test was passing before. Why are we skipping it?

internal/provider/utils_test.go
```diff
@@ -1,12 +1,121 @@
...
+		{
+			id:          types.Int64Null(),
+			name:        types.StringNull(),
+			orgName:     types.StringValue("org1"),
+			URI:         "inventories",
+			expectError: errors.New("invalid lookup parameters"),
+			expectedUrl: "",
+		},
+	}
+	for _, test := range testTable {
```
dleehr, line 106: 👍 Nice testing table

dleehr (commented):
Thanks for running the tests and posting the output. Excited to see these enhancements coming 😄. Overall I think this looks really good!

After my last review we had some good discussion around some docs suggestions (required->optional, reference for`named_url`). I thought we had a good direction from the discussion with some small changes. I see the comments are now resolved but there were no code changes. I don't want this to get lost, is there a plan to address it?

My other feedback was around the acceptance tests. I see one has been updated to verify the new attributes are present, so that's great 👍. However, one that was active is now being skipped (which seems like a regression). Most importantly, the use case we're aiming to support (being able to use a `data.aap_inventory` to look up an inventory by name) is not covered by an acceptance test at all. Is that a test we can write?

matoval (approved):
lgtm

dleehr:
> Acceptance tests were updated, and now docs updated so that should cover everything that has been discussed.

OK I see that an acceptance test for the inventory resource was updated to check for the new attributes, but I do not see any **new** acceptance tests around the new functionality described in the story:

> As a user of Terraform and AAP, I want the [terraform-provider-aap](https://registry.terraform.io/providers/ansible/aap) to implement [data sources](https://developer.hashicorp.com/terraform/language/data-sources) for the Job Template and Inventory objects in AAP so that I can look up existing objects when creating AAP jobs from Terraform.

It's good that we're testing these changes (no regressions 🎉) but we need new acceptance tests that follow the user story - show that we can look up an inventory by `name` and not just `id`:

> more importantly I think we need to add acceptance tests for this functionality

> Most importantly, the use case we're aiming to support (being able to use a data.aap_inventory to look up an inventory by name) is not covered by an acceptance test at all. Is that a test we can write?

The Jira has a pretty basic terraform config we should be able to use in an acceptance test to confirm. Maybe we need organization name too, but generally we'd use something like this, right?

```terraform
data "aap_inventory" "my_inventory" {
  name = "my_inventory" # no id
}
```

(P.S. can we check if we're really addressing #52 with these changes? I'm not sure this PR closes that issue. 👀 )

dleehr (approved):
Discussed and agree that acceptance tests are needed for the story to be complete but not to hold up this PR.

PabloHiro (approved):
Let's get this in and follow up on https://github.com/ansible/terraform-provider-aap/pull/64

PR #63, opened 2025-05-06
Fix golangci-lint configuration

lranjbar:


AaronH88 (approved):
lgtm

arrestle:
Ah I see this is for migration: https://golangci-lint.run/product/migration-guide/ (note I've not gone into depth) but I believe we may want to [keep linting on shadow variables](https://medium.com/@shahpershahin/variable-shadowing-in-go-best-practices-to-avoid-confusions-and-bugs-61e03022b54d) while upgrading our golangci-lint version

```
  enable:
    - govet

linters-settings:
  govet:
    enable:
      - shadow
```

dleehr:
> Just a question, is this causing some sort of problem? Is this for a migration to a later version of golangci-lint?

Yeah, our other PRs are failing the lint step because of the config. See https://github.com/ansible/terraform-provider-aap/actions/runs/14886287366/job/41806639015.

This one is passing https://github.com/ansible/terraform-provider-aap/actions/runs/14871486836/job/41760644297?pr=63

dleehr (approved):
Thanks for the fix! I noticed my PR #61 is failing the lint step due to a bad config so I came to the list of PRs to see if any of them were passing and found this 🙌

arrestle:
> > Just a question, is this causing some sort of problem? Is this for a migration to a later version of golangci-lint?
> > Ah I see this is for migration: https://golangci-lint.run/product/migration-guide/ (note I've not gone into depth) but I believe we may want to [keep linting on shadow variables](https://medium.com/@shahpershahin/variable-shadowing-in-go-best-practices-to-avoid-confusions-and-bugs-61e03022b54d) while upgrading our golangci-lint version
> > ```
> >   enable:
> >     - govet
> > 
> > linters-settings:
> >   govet:
> >     enable:
> >       - shadow
> > ```
> 
> Sure I can add it back into the settings. It was just failing with incorrect settings so I went with the easiest route. I'll try to update it. We also need to update golangci-lint to 2.x which will require more config changes. 🤔 Same for receptor

Didn't realize it was causing pipelines to fail. Sheesh. Yeah. Take it out!

PR #66, opened 2025-05-14
Disabling dupl linter due to false positives

lranjbar:
The provider is using a lot of boilerplate code. This linter is currently picking up on functions that are duplicate but are apart of the boilerplate. In the background we are looking at ways to de duplicate the code. In the meantime lets disable the linter.


matoval (approved):
lgtm

arrestle (approved):
lgtm

PR #69, opened 2025-05-15
Reduce technical debt of code duplication in automated tests

arrestle:
1. Updates to remove duplicate code and share constants and variables.
2. Updates to the Makefile to add help, allow local linting through local .tools folder, remove hardcoded AAP_HOST.


internal/provider/utils_test.go
```diff
@@ -2,12 +2,79 @@ package provider
...
+}
+
+func checkBasicInventoryAttributes(t *testing.T, name, expectedName string) resource.TestCheckFunc {
+	t.Helper()
+	return resource.ComposeAggregateTestCheckFunc(
+		resource.TestCheckResourceAttr(name, "name", expectedName),
+		resource.TestCheckResourceAttr(name, "organization", "1"),
+		resource.TestCheckResourceAttr(name, "organization_name", "Default"),
+		resource.TestMatchResourceAttr(name, "url", reInventoryURLPattern),
+		resource.TestCheckResourceAttr(name, "named_url", fmt.Sprintf("/api/controller/v2/inventories/%s++%s/", expectedName, "Default")),
```
dleehr, line 72: I think this will fail on 2.4 where we don't have `/controller` in the URL. I know it's not new code, but we didn't catch it in the change that introduced the `named_url` attribute.

For the rest of the checks we have `(/controller)?/` in the RegExp patterns so that these should succeed on both 2.4 and 2.5.

Honestly I think there's value in leaving it as-is. When we get pipelines up and running to test this on 2.4 and 2.5 that should catch this. The test is passing, so I'm not requesting a change here.

dleehr (changes requested):
Thanks for getting this in so early @arrestle.

I've skimmed the refactoring in the `_test.go` and utils files, and that looks good. Just a couple of points there. I also ran the acceptance tests locally before/after and confirmed the test coverage has not decreased. I would appreciate if others could weigh in as well on the refactored logic.

I'm requesting changes on two specific areas:

1. There are changes in the README and Makefile around what to run to support the acceptance tests. I think we'll be addressing that in a future ticket around contribution guidelines, so I don't think we're ready for this change and my request is that we hold these back.
2. There are additions to `GetWithStatus` that enhance error checking. I had some specific feedback on individual checks as well and some general refactoring feedback. The ask here is to review these checks and see what can be refactored, removed, or changed so that details aren't lost.

internal/provider/client.go
```diff
@@ -150,8 +153,28 @@ func (c *AAPClient) Create(path string, data io.Reader) ([]byte, diag.Diagnostic
...
 func (c *AAPClient) GetWithStatus(path string) ([]byte, diag.Diagnostics, int) {
+	if c == nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("AAPClient is nil", "Client was not initialized properly")}, 0
+	}
+	if path == "" {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("Path is empty", "Path cannot be empty")}, 0
+	}
+	if c.httpClient == nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("HttpClient is nil", "HttpClient was not initialized properly")}, 0
+	}
```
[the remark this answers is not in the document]
arrestle: I mean't to remove these changes. This is pushed to  another ticket.

internal/provider/client.go
```diff
@@ -150,8 +153,28 @@ func (c *AAPClient) Create(path string, data io.Reader) ([]byte, diag.Diagnostic
...
 	getResponse, body, err := c.doRequest("GET", path, nil)
 	diags := ValidateResponse(getResponse, body, err, []int{http.StatusOK})
+	if diags.HasError() {
+		return nil, diags, getResponse.StatusCode
+	}
+	if getResponse == nil {
+		diags.AddError("Get response is nil", "The response from the server is nil")
+		return nil, diags, 0
+	}
```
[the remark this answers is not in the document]
arrestle: You originally asked me to put this in the backlog, but I do have a pr out https://github.com/ansible/terraform-provider-aap/pull/70 to handle it. I'm fine with rolling it in here.

internal/provider/client.go
```diff
@@ -150,8 +153,28 @@ func (c *AAPClient) Create(path string, data io.Reader) ([]byte, diag.Diagnostic
...
+	if body == nil {
+		diags.AddError("Get response body is nil", "The response body from the server is nil")
+		return nil, diags, 0
+	}
```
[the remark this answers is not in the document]
arrestle: yes meant to take these out.

internal/provider/utils_test.go
arrestle: Let me know if I need to make any changes here.

dleehr: No, I think we can leave this as-is. This test was already on `main` this way and I expect it fails on `awx` and AAP 2.4, so I think there's value in seeing it fail on 2.4 and then fixing it. Thanks!

Makefile
```diff
@@ -1,32 +1,33 @@
...
-test:
+test: ## Execute all unit tests with verbose output."
 	@echo "==> Running unit tests..."
 	go test -v ./...
 
-testacc:
+testacc: ## Run acceptance tests against instance (https://localhost:8043)."
 	@echo "==> Running acceptance tests..."
-	TF_ACC=1 AAP_HOST="https://localhost:8043" AAP_INSECURE_SKIP_VERIFY=true go test -count=1 -v ./...
+	TF_ACC=1 AAP_HOST="https://localhost:8043" go test -count=1 -v ./...
```
dleehr, line 18:
```suggestion
	TF_ACC=1 AAP_HOST="https://localhost:8043" AAP_INSECURE_SKIP_VERIFY=true go test -count=1 -v ./...
```

internal/provider/utils_test.go
```diff
@@ -2,12 +2,79 @@ package provider
...
+func checkBasicGroupAttributes(t *testing.T, name, expectedName string) resource.TestCheckFunc {
+	t.Helper()
+	return resource.ComposeAggregateTestCheckFunc(
+		resource.TestCheckResourceAttr(name, "name", expectedName),
+		resource.TestCheckResourceAttrPair(name, "inventory_id", resourceNameInventory, "id"),
+		resource.TestMatchResourceAttr(name, "url", reGroupURLPattern),
+	)
+}
+
+func checkBasicInventoryAttributes(t *testing.T, name, expectedName string, expected_orgId string, expected_orgName string) resource.TestCheckFunc {
```
dleehr, line 65: I think I commented on the commit directly by mistake so apologies if this is a dupe

nit: The variable names should use a consistent convention - `expectedOrgId` and `expectedOrgName`

README.md
```diff
@@ -44,9 +44,12 @@ Acceptance tests apply test terraform configurations to a running AAP instance a
 
 To run acceptance tests locally, start a local AAP instance following the [docker-compose instructions for local AWX development](https://github.com/ansible/awx/blob/devel/tools/docker-compose/README.md). Create an admin user for the AAP instance and save the credentials to these environment variables:
 
+Create an admin user for the AAP instance and set the following environment variables:
```
tyraziel, line 47: I don't think the user needs to be an admin - that would get them the right level of permissions, but I think we should try to go with least privilege required.  On the other hand, if the lease privilege is admin, ignore my comment and request for changes :)

Makefile
tyraziel, line 1: I'd almost split out the Makefile changes into a separate PR.

makefiles/golangci.mk
tyraziel, line 1: If the main makefile gets put into its own PR, don't forget this sub makefile.

tyraziel (changes requested):

internal/provider/client.go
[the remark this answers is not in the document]
dleehr: Backlog is fine.

makefiles/golangci.mk
```diff
@@ -0,0 +1,24 @@
+
+ifndef GOLANGCI_LINT_MK_INCLUDED
+
+TOOLS_DIR := $(CURDIR)/.tools
```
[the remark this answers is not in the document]
arrestle: There wasn't a make clean originally. Adding now.

makefiles/golangci.mk
arrestle: We talked about this on [slack](https://redhat-internal.slack.com/archives/C08CJ87JRG9/p1747328677721189) but noone came out and asked for a different ticket then. If asked I'm happy to find a ticket, write a ticket or just add a second pr to this ticket. Just say the word.

Makefile
```diff
@@ -1,32 +1,33 @@
...
-	@echo "==> Checking source code against linters..."
-	golangci-lint run -v ./...
-
-test:
+test: ## Execute all unit tests with verbose output."
 	@echo "==> Running unit tests..."
 	go test -v ./...
 
-testacc:
+testacc: ## Run acceptance tests against instance (https://localhost:8043)."
```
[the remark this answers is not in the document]
arrestle: I agree. I put all the environment variables into the README.md and removed reference to aap-dev.

README.md
```diff
@@ -44,9 +44,12 @@ Acceptance tests apply test terraform configurations to a running AAP instance a
...
+
 ```bash
 export AAP_USERNAME=<your admin username>
 export AAP_PASSWORD=<your admin password>
+export AAP_HOST="http://localhost:9080" # if using aap-dev (Note: Subject to change)
```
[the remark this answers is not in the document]
arrestle: Lots of comments here. Came up with:

`export AAP_HOST=<your aap instance host url> # "http://localhost:9080" or "https://localhost:8043" `

dleehr (approved):
Thanks for addressing the feedback. LGTM.

PR #72, opened 2025-05-29
Changelog and documentation updates to prep for release

dleehr:
- Adds a changelog fragment with entries for the changes since the last release
- Updates documentation and examples @tyraziel 
- Fixes a helper function in `test_utils.go` that caused a failure on AAP 2.4
- Acceptance tests have been run against AAP 2.4 and 2.5 instances, with logs uploaded to the issue tracker.


docs/resources/inventory.md
```diff
@@ -114,29 +114,3 @@ output "inventory_xyz" {
...
-
-Creating a new inventory in the Default organization:
-```terraform
-resource "aap_inventory" "sample" {
-  name        = "My Sample Inventory"
-  organization_name = "Default"
-  description = "A new inventory for testing"
-  variables   = jsonencode(yamldecode(local.values_variables))
-}
-```
```
dleehr, line 130: I tested this with a local build of the provider off `main` and `terraform plan` fails:

```
│ Error: Invalid Configuration for Read-Only Attribute
│ 
│   with aap_inventory.sample,
│   on main.tf line 37, in resource "aap_inventory" "sample":
│   37:   organization_name = "Default"
│ 
│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value.
│ 
│ Refer to the provider documentation or contact the provider developers for additional information about configurable and read-only attributes
│ that are supported.
```

Since the resource doesn't support `organization_name`, I reorganized this to highlight that in the data source docs, where it is supported.

dleehr: > This is what I was thinking about with my comment above, we should fix the documentation. If anything we can provide examples of how to get the organization by name and use that in this example.

Agreed. And if it wasn't clear, that's what the change does :)

PR #77, opened 2025-06-04
Refactor Data Source Model classes fields and methods out to a parent struct

fincamd:
This new struct (BaseDataSource) now contains the shared methods by all the different *DataSource structs.

We also provide a BaseDataSourceModel struct that does the same thing with the shared model attributes and ParseHttpResponse method.


fincamd:
Closed in favour of #81

PR #81, opened 2025-06-09
Add BaseDataSourceModel and BaseDataSource structs with refactored functions

fincamd:
These functions come from the different DataSource entities defined in other files. We saw an opportunity to refactor these repeated functions and common fields out to a new base struct.

This is the first part of the refactoring effort. There will be other PRs linked to this one later on removing the corresponding refactored code.

UPDATE:
In the `general_api_models.go` file you will find three different inheritance trees. All of these support the variants with and without AAP organization. So that we can better match the API's schemas to the Go struct fields.

- `BaseDetailAPIModel`, `BaseDetailAPIModelWithOrg`, and children/composed structs. These represent AAP's API fields returned in `Read` requests, that get parsed into the following struct hierarchy.
- `BaseDetailDataSourceModel`, and children `BaseDetailDataSourceModelWithOrg` that will receive the parsed information from incoming requests of the Read operation.
- `BaseDataSource`, and `BaseDataSourceWithOrg`. These incorporate all the required functions of the `datasource.DataSource` interface of the provider SDK. Some functions like `Read` or `ConfigValidators` have been overloaded to support the inheritance trees of with/without AAP organization. Due to Go's static typing, this refactor isn't possible other way without smashing every field together into a single struct as it was before.

With this refactor, this is an example of a DataSource file. 31 lines!
```go
package provider

import (
	"github.com/hashicorp/terraform-plugin-framework/datasource"
)

// JobTemplate AAP API model
type JobTemplateAPIModel struct {
	BaseDetailAPIModelWithOrg
}

// JobTemplateDataSourceModel maps the data source schema data.
type JobTemplateDataSourceModel struct {
	BaseDetailDataSourceModelWithOrg
}

// JobTemplateDataSource is the data source implementation.
type JobTemplateDataSource struct {
	BaseDataSourceWithOrg
}

// NewJobTemplateDataSource is a helper function to simplify the provider implementation.
func NewJobTemplateDataSource() datasource.DataSource {
	return &JobTemplateDataSource{
		BaseDataSourceWithOrg: *NewBaseDataSourceWithOrg(nil, StringDescriptions{
			MetadataEntitySlug:    "job_template",
			DescriptiveEntityName: "JobTemplate",
			ApiEntitySlug:         "job_templates",
		}),
	}
}
```

### DIFF
--- a/internal/provider/general_api_models.go
+++ b/internal/provider/general_api_models.go
@@ -1,5 +1,20 @@
 package provider
 
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"path"
+
+	"github.com/ansible/terraform-provider-aap/internal/provider/customtypes"
+	"github.com/hashicorp/terraform-plugin-framework-validators/datasourcevalidator"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	tfpath "github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
 type SummaryFieldsAPIModel struct {
 	Organization SummaryAPIModel `json:"organization,omitempty"`
 	Inventory    SummaryAPIModel `json:"inventory,omitempty"`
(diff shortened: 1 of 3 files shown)


internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+	uri := path.Join(d.client.getApiEndpoint(), d.ApiEntitySlug)
+	resourceURL, err := ReturnAAPNamedURL(state.Id, state.Name, state.OrganizationName, uri)
+	if err != nil {
+		resp.Diagnostics.AddError("Minimal Data Not Supplied", "Expected either [id] or [name + organization_name] pair")
+		return
+	}
+
+	readResponseBody, diags := d.client.Get(resourceURL)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
```
AaronH88, line 117: Same comment for the next 3 if conditions, we just return, should we generate an error?

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+func (d *BaseDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if req.ProviderData == nil {
+		return
+	}
+
+	client, ok := req.ProviderData.(*AAPClient)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected Data Source Configure Type",
+			fmt.Sprintf("Expected *AAPClient, got: %T. Please report this issue to the provider developers.", req.ProviderData),
```
AaronH88, line 144: We dont know what `req.ProviderData` is but we attempt to print it, Does this open us up for a log injection attack? Should we attempt to `Sanitize` this `req.ProviderData` before printing, I think we should

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+	// Set state
+	// Save updated data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Configure adds the provider configured client to the data source.
+func (d *BaseDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
```
AaronH88, line 135: I think I have raised this before but I can not remember the outcome, we do not check `context`'s to ensure they are still valid. Is this a conscious decisions from the team or tec debt?

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+// and an apiEntitySlug string indicating the entity path name to consult the API.
+func NewBaseDataSource(client ProviderHTTPClient, stringDescriptions StringDescriptions) *BaseDataSource {
+	return &BaseDataSource{
+		client:             client,
+		StringDescriptions: stringDescriptions,
+	}
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (d *BaseDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
```
AaronH88, line 98: `ctx` is passed along but never checked, I think we could check it and fail/return if the `ctx` is `Done`

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+		),
+	}
+}
+
+func (d *BaseDataSource) ValidateConfig(ctx context.Context, req datasource.ValidateConfigRequest, resp *datasource.ValidateConfigResponse) {
+	var data BaseDataSourceModel
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
```
AaronH88, line 171: print/create error before returning?

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+		resp.Diagnostics.AddAttributeWarning(
+			tfpath.Root("name"),
+			"Missing Attribute Configuration",
+			"Expected name to be configured with organization_name.",
+		)
+	}
+}
+
+// Schema defines the schema fields for the data source.
+func (d *BaseDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
AaronH88, line 209: General comment about contexts 😄

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+	var state BaseDataSourceModel
+	var diags diag.Diagnostics
+
+	// Read Terraform configuration data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	uri := path.Join(d.client.getApiEndpoint(), d.ApiEntitySlug)
```
AaronH88, line 108: The constructor states that the `client`s default is nil, 

> client instance (usually initialized to nil, it will be later configured calling the Configure function)

Lets check to ensure the client exists before trying to call a function as we might get a problem here.

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (d *BaseDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var state BaseDataSourceModel
+	var diags diag.Diagnostics
+
+	// Read Terraform configuration data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
```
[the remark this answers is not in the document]
fincamd: I don't know whether this is being handled elsewhere. I'll look around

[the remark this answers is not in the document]
fincamd: ## TL; DR
Reading from the documentation, we shouldn't do other thing than update the Diagnostics object with warnings or errors. Then Terraform will worry about letting the user know about those problems. Plus, the state is still updated even with errors present.

---

As per this page of docs: https://developer.hashicorp.com/terraform/plugin/framework/diagnostics, the providers should use the `Diagnostics` object provided by the `diag` package to report errors/warnings etc.:

> Returning errors and warnings
> Providers use Diagnostics to surface errors and warnings to practitioners, such as contextual messages returned from Terraform CLI at the end of command output
> [...]
> This is the most common form for Diagnostics: a slice that has one or more errors appended to it. This approach allows your provider to inform practitioners about all relevant errors and warnings at the same time, allowing practitioners to fix their configuration or environment more quickly. You should only append to Diagnostics slices and never replace or remove information from them.

It seems this is the best practice there is when reacting to errors. I understand from this that we shouldn't be raising any extra errors or halting the execution. This also takes into account another section of the documentation that states:

> How Errors Affect State
> Returning an error diagnostic does not stop the state from being updated. Terraform will still persist the returned state even when an error diagnostic is returned with it. This is to allow Terraform to persist the values that have already been modified when a resource modification requires multiple API requests or an API request fails after an earlier one succeeded.
>
> When returning error diagnostics, we recommend resetting the state in the response to the prior state available in the configuration.

Hence, if we stopped the execution by raising some type of RuntimeError or the equivalent in Go, the state wouldn't be updated and some users could be affected by this.

Too long a comment, WDYT??

internal/provider/general_api_models.go
fincamd: [My previous comment](https://github.com/ansible/terraform-provider-aap/pull/81#discussion_r2135571529) can work here too.

In this specific case, I don't think we should do other thing than return from the current function. It doesn't sit well with me, but from what I gather from the documentation the providers should do as much as possible (gathering errors and warnings from all the operations), then let the user know about all the issues. 

Otherwise, if we halted the execution by raising an error the user would take longer to fix the issues. Errors/Warnings printing is taken care of by the terraform `diag` model when Terraform runs.

internal/provider/general_api_models.go
AaronH88: Also I just found out that we shouldnt be using fmt.Sprintf for logging, link: https://developer.hashicorp.com/terraform/plugin/log/writing

internal/provider/general_api_models.go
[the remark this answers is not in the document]
fincamd: >What happen if req.Config is nil,

That definitely is a failing scenario. Should we just add a new diagnostic saying that we couldn't load config? Or at least try to run the Config() function again then retry? It seems like a corner case that the req.Config field is nil.

If I follow the purpose of that field correctly, it contains the properties + values that the user configured in the data source HCL. So if it was nil, it wouldn't be our fault or the user's, I believe it'd be a problem with the Terraform parser. Do we trust Terraform to always populate that value even though it could have nil values inside?

I am happy to add defensive programming for these cases too. I'm just thinking outloud.

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Configure adds the provider configured client to the data source.
+func (d *BaseDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if req.ProviderData == nil {
+		return
```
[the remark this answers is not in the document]
fincamd: I'll add this. Thanks for the suggestion!!

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+	}
+
+	client, ok := req.ProviderData.(*AAPClient)
+	if !ok {
+		resp.Diagnostics.AddError(
```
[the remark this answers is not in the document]
fincamd: Makes sense, the provider must be full of these lines. We might then find an opportunity to create a reusable function to check this for us in the future.

I'll add the change for now. Thanks!

internal/provider/general_api_models.go
fincamd: This sounds like a big change for the provider. There are a lot of places where we are currently ignoring the contexts. I guess this can be part of a big enhancement effort

internal/provider/general_api_models.go
fincamd: Resolving this conversation since the terraform SDK will print any Diagnostics as part of its execution

internal/provider/general_api_models.go
```diff
@@ -16,3 +31,227 @@ type SummaryAPIModel struct {
...
+					"variables were entered into AAP.", d.DescriptiveEntityName),
+			},
+		},
+		Description: fmt.Sprintf("Get an existing %s.", d.DescriptiveEntityName),
+	}
+}
+
+// Metadata returns the data source type name composing it from the provider type name and the
+// entity slug string passed in the constructor.
+func (d *BaseDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
```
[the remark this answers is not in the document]
fincamd: I have added the `resp` nil pointer check. The one for the `req` is not needed since it is a struct type. The one about the context, we'd need the Diagnostics object and the response doesn't have it. Plus, the function would take instants to run since it's only doing String manipulations. Do we think this is enough?

dleehr (changes requested):
Thanks for the PR. Looks like valuable refactoring foundation work and appreciate addressing feedback from the team to make it more robust.

I'm requesting changes specifically around `ParseHttpResponse`, because what's here binds the base implementation to the specific `WorkflowJobTemplateAPIModel` API Model struct, and it's not clear to me how that will support other data sources that represent other AAP objects.

So specifically what I'd like to see clarified is: How do we expect this `BaseDataSourceModel` to be used when implementing other data sources? Is there some other to make this more generic (introspection?) Do developers need to override `ParseHttpResponse` with their own implementation? If so, perhaps the base implementation should fail with some kind of "Not Implemented" error.

internal/provider/general_api_models.go
```diff
@@ -16,3 +32,299 @@ type SummaryAPIModel struct {
...
+type BaseDataSourceModel struct {
+	Id               types.Int64                      `tfsdk:"id"`
+	Name             types.String                     `tfsdk:"name"`
+	Organization     types.Int64                      `tfsdk:"organization"`
+	OrganizationName types.String                     `tfsdk:"organization_name"`
+	Url              types.String                     `tfsdk:"url"`
+	NamedUrl         types.String                     `tfsdk:"named_url"`
+	Description      types.String                     `tfsdk:"description"`
+	Variables        customtypes.AAPCustomStringValue `tfsdk:"variables"`
+}
```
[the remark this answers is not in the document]
fincamd: Thanks for the suggestion @PabloHiro. Done in [this commit](https://github.com/ansible/terraform-provider-aap/pull/81/commits/1c31d877395561f976c5be88dd66a56d88467134)

dleehr (approved):
LGTM. The questions I had have been addressed. Thanks!

internal/provider/general_api_models.go
```diff
@@ -16,3 +32,324 @@ type SummaryAPIModel struct {
 type RelatedAPIModel struct {
 	NamedUrl string `json:"named_url,omitempty"`
 }
+
+// A base struct for the entities API models. To be extended as needed.
+type BaseEntityAPIModel struct {
```
PabloHiro, line 37: Have been thinking about this one for a while. Can we have this instead?
```
// A base struct for the Detail API view. To be extended as needed.
type BaseDetailAPIModel struct {
   Id          int64  `json:"id"`
   Name string `json:"name"`
   Description string `json:"description,omitempty"`
   SummaryFields SummaryAPIModel `json:"summary_fields"`
   Url string `json:"url"`
   Related RelatedAPIModel `json:"related"`	
   Organization int64 `json:"organization,omitempty"`
   Inventory int64 `json:"inventory,omitempty"`
   Variables       string          `json:"variables,omitempty"`
}
```

Reasoning:
* Use the word `Detail` instead of `Entity` to refer to Detail views in Django Rest Framework, which is what AAP is built from.
* Have a similar modeling structure here to what is present in https://github.com/ansible/django-ansible-base
* Embedding `SummaryAPIModel` to make the Id/Name/Description fields available, but also having a `summary_fields` in the JSON payload induces confusion. Make `SummaryAPIModel` available under `SummaryFields`
* Remove `omitempty` from the truly base fields present in all API responses.
* `organization` and `inventory`, if present at the root are an int64. Only when nested under summary_fields is when they have fields.

PabloHiro (changes requested):

internal/provider/general_api_models.go
PabloHiro: Another option is to remove `Description`, `Organization`, `Inventory` and `Variables` from here and define them explicitly in the models representing real resources.

PabloHiro: I encourage you to take a look at different API responses in an AAP instance, and then re-review the API models we have in this repo.

internal/provider/general_api_models.go
```diff
@@ -16,3 +29,477 @@ type SummaryAPIModel struct {
...
+	}
+
+	// TODO: REVIEWERS ABOUT THESE LINES: Should this function do something else when the entity doesn't support organization?
+	// This function makes sense to me if the entity doesn't support organization
+
+	if !IsValueProvided(data.Id) && !IsValueProvided(data.Name) {
+		resp.Diagnostics.AddAttributeWarning(
+			tfpath.Root("id"),
+			"Missing Attribute Configuration",
+			"Expected either [id]",
```
PabloHiro, line 377: W can get rid of these lines. Taking a look at:
https://developer.hashicorp.com/terraform/plugin/framework/resources/validate-configuration

The validation we want to perform is already in `ConfigValidators`

internal/provider/general_api_models.go
```diff
@@ -16,3 +29,477 @@ type SummaryAPIModel struct {
...
+			"description": schema.StringAttribute{
+				Computed:    true,
+				Description: fmt.Sprintf("Description of the %s", d.DescriptiveEntityName),
+			},
+			"variables": schema.StringAttribute{
+				Computed:   true,
+				CustomType: customtypes.AAPCustomStringType{},
+				Description: fmt.Sprintf("Variables of the %s. Will be either JSON or YAML string depending on how the "+
+					"variables were entered into AAP.", d.DescriptiveEntityName),
+			},
```
PabloHiro, line 484:
```suggestion
```
Obsolete, other fields here as well you might want to get rid of.

internal/provider/general_api_models.go
```diff
@@ -1,12 +1,25 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
 
-// If we end up pulling in more summary_fields that have other information, we can split
-// them out to their own structs at that time.
+// ---------------------------------------------------------------------------
+
+// TODO: Delete this once the PR of the refactor is merged
+// This struct is just here to make the CI happy :D
```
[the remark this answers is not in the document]
fincamd: I mean, that's going to disappear once I apply the refactor. It's only there to make the CI happy.
I had to break down the changes so reviewers would have an easier time. This change to add Deprecation will not be needed. Thanks!

internal/provider/general_api_models.go
```diff
@@ -16,3 +29,477 @@ type SummaryAPIModel struct {
...
+		return
+	}
+
+	// Read Terraform configuration data into the model
+	var state BaseDetailDataSourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &state)...)
+	uri := path.Join(d.client.getApiEndpoint(), d.ApiEntitySlug)
+
+	// TODO: REVIEWERS ABOUT THESE LINES: Should this NamedUrl be created when the entity doesn't support organization?
+	resourceURL, err := ReturnAAPNamedURL(state.Id, state.Name, types.StringValue(""), uri)
```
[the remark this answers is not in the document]
fincamd: > Maybe we would need ReturnAAPNamedURL and ReturnAAPNamedURLWithOrg, and do not pass that field here.

Another option would be to write a different function that does what you suggest, but then, we'd be duplicating code and/or needing to move that empty string or null string elsewhere

internal/provider/general_api_models.go
```diff
@@ -16,3 +29,477 @@ type SummaryAPIModel struct {
...
+	// TODO: REVIEWERS ABOUT THESE LINES: Should this ConfigValidators be created when the entity doesn't support organization?
+
+	// You have at least an id or a name + organization_name pair
+	return []datasource.ConfigValidator{
+		datasourcevalidator.Any(
+			datasourcevalidator.AtLeastOneOf(
+				tfpath.MatchRoot("id")),
+			datasourcevalidator.RequiredTogether(
+				tfpath.MatchRoot("name"),
+				tfpath.MatchRoot("organization_name")),
```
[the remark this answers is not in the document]
fincamd: Added in [this commit](https://github.com/ansible/terraform-provider-aap/pull/81/commits/a6c8bc7f4d5b869561428a2a951902f82c523f6e)

internal/provider/general_api_models.go
```diff
@@ -16,3 +29,477 @@ type SummaryAPIModel struct {
...
+// A struct to represent a base DataSource object, with a client and the slug name of
+// the API entity.
+type BaseDataSource struct {
+	client ProviderHTTPClient
+	StringDescriptions
+}
+
+type BaseDataSourceWithOrg struct {
+	BaseDataSource
+}
```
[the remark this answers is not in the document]
fincamd: >Shouldn't these be used as parents of BaseDetailDataSourceModel and BaseDetailDataSourceModelWithOrg?

No, because these are different structs. One represents the state of the data source. The other is what we use to implement the interface defined for datasource.DataSource entities of the SDK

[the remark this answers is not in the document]
fincamd: >I would also define these before BaseDetailDataSourceModel and BaseDetailDataSourceModelWithOrg. That way someone reading the file does not need to scroll down and up.

Let me add that as a new commit

PabloHiro (approved):
LGTM, we can refine the almost duplicated `Read` in follow up PRs.

internal/provider/general_api_models.go
fincamd: I have added the deprecation message to the field

PR #82, opened 2025-06-09
Update to Golang 1.23.9

lranjbar:
Updates Golang version to 1.23.9


tyraziel (approved):
lgtm

.github/workflows/unit_test_and_quality.yml
```diff
@@ -62,11 +62,11 @@ jobs:
 
       # - name: Run Tests with Coverage
       #   run: make testcov
-      
+
```
tyraziel, line 65:
```suggestion
```

I originally left these, and since we're modifying this file, let's just remove them.

tyraziel (approved):
Re-approving.

PR #84, opened 2025-06-13
SonarCloud Integration

tyraziel:
Integrating with SonarCloud

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,10 @@ build: ## Compile the Go package. This is the default.
 	go build
 
 clean: ## Remove the .tools folder and binaries.
-	@rm -rf .tools
-	@rm terraform-provider-aap || true
-	@rm unit-testing.cov || true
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -11,16 +11,26 @@ build: ## Compile the Go package. This is the default.
 
 clean: ## Remove the .tools folder and binaries.
 	@rm -rf .tools
-	@rm terraform-provider-aap
+	@rm terraform-provider-aap || true
```
schmidtd, line 14: À propos of nothing in particular, you can ignore results in Make with a minus in front of the production:
```
-@rm terraform-provider-aap
```

tyraziel: Thanks for that!  I'll adopt and update.

PabloHiro (approved):
LGTM, since this happens on main and it is hard to test from a fork. I'd say merge this now and open follow up PRs to adjust.

PR #93, opened 2025-06-23
Running code coverage tests within the Sonar Cloud Static Scan action

tyraziel:
Removed the code coverage generation from the main unit testing action, and placed it in the sonar scan action.

### DIFF
--- a/.github/workflows/sonar-checks.yml
+++ b/.github/workflows/sonar-checks.yml
@@ -11,14 +11,30 @@ jobs:
   sonarcloud:
     name: Sonar Cloud Static Scans
     runs-on: ubuntu-latest
+    env:
+        go_version: '1.23'
+        
     if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request'
 
     steps:
-      - name: Fetch Code Coverage Report
-        uses: actions/download-artifact@v4
+      - name: Checkout Provider
+        uses: actions/checkout@v4
         with:
-          name: coverage-report
-          path: .
+          fetch-depth: 0
+
+      - name: Setup Go ${{ env.go_version }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.go_version }}
+
+      - name: Run unit tests with coverage
+        run: make testcov
+
+      # - name: Fetch Code Coverage Report
+      #   uses: actions/download-artifact@v4
+      #   with:
+      #     name: coverage-report
+      #     path: .
 
       - name: Sonar Cloud Scan
         uses: SonarSource/sonarqube-scan-action@v5.2.0
--- a/.github/workflows/unit_test_and_quality.yml
+++ b/.github/workflows/unit_test_and_quality.yml
@@ -30,11 +30,14 @@ jobs:
       - name: Build the Provider
         run: go build -v ./...
 
-      - name: Run unit tests with coverage
-        run: make testcov
+      - name: Run Provider Unit Tests
+        run: make test
 
-      - name: Upload code coverage report from unit tests
-        uses: actions/upload-artifact@v4
-        with:
-          name: coverage-report
-          path: unit-testing.cov
+      # - name: Run unit tests with coverage
+      #   run: make testcov
+
+      # - name: Upload code coverage report from unit tests
+      #   uses: actions/upload-artifact@v4
+      #   with:
+      #     name: coverage-report
+      #     path: unit-testing.cov


.github/workflows/sonar-checks.yml
```diff
@@ -11,14 +11,30 @@ jobs:
...
+          go-version: ${{ env.go_version }}
+
+      - name: Run unit tests with coverage
+        run: make testcov
+
+      # - name: Fetch Code Coverage Report
+      #   uses: actions/download-artifact@v4
+      #   with:
+      #     name: coverage-report
+      #     path: .
```
dleehr, line 37: Let's delete these rather than leaving them commented-out.

.github/workflows/unit_test_and_quality.yml
```diff
@@ -30,11 +30,14 @@ jobs:
...
-          name: coverage-report
-          path: unit-testing.cov
+      # - name: Run unit tests with coverage
+      #   run: make testcov
+
+      # - name: Upload code coverage report from unit tests
+      #   uses: actions/upload-artifact@v4
+      #   with:
+      #     name: coverage-report
+      #     path: unit-testing.cov
```
dleehr, line 43: Same feedback as above.

.github/workflows/sonar-checks.yml
```diff
@@ -11,14 +11,30 @@ jobs:
...
-          path: .
+          fetch-depth: 0
+
+      - name: Setup Go ${{ env.go_version }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.go_version }}
+
+      - name: Run unit tests with coverage
+        run: make testcov
```
dleehr, line 31: The `unit_test_and_quality.yml` workflow where we're moving this from has more steps before this (`go get`, `go build`. Do we need those steps here? If we don't, why do we have them in the other workflow?

dleehr (changes requested):
Requesting changes because I don't understand the discrepancy in steps and we should delete the commented-out code if it's not needed.

Broader question - we've now merged 4 PRs so far to configure sonar scans. How confident are we that the scans will be working after this? 

I know it's difficult to test these changes without getting the code onto `main`, but if the scans/coverage don't succeed after these changes, I would like to see this working in a sandbox repo (where you can PR/merge without approval) before bringing a PR back.

.github/workflows/unit_test_and_quality.yml
tyraziel: I'd like to leave this in, as at some point we should try to get this working, instead of trying to run the code coverage tests again.

.github/workflows/sonar-checks.yml
tyraziel: If something fails in those discrete steps during the initial PR workflow, it's a bit easier to see and know what's failing.  Once we get here, I don't see a point of splitting it all out again.  This is a stop gap measure to get this working without trying to submit more PRs to get the upload / download of the code coverage working.

tyraziel:
> Requesting changes because I don't understand the discrepancy in steps and we should delete the commented-out code if it's not needed.
> 
> Broader question - we've now merged 4 PRs so far to configure sonar scans. How confident are we that the scans will be working after this?
> 
> I know it's difficult to test these changes without getting the code onto `main`, but if the scans/coverage don't succeed after these changes, I would like to see this working in a sandbox repo (where you can PR/merge without approval) before bringing a PR back.

Unfortunately, I think this is how we have to work when we're making changes to github actions / workflows and CI, so there will always be some level of uncertainty until we run the workflows the way they are intended to run.

PabloHiro (approved):

dleehr:
> Unfortunately, I think this is how we have to work when we're making changes to github actions / workflows and CI, so there will always be some level of uncertainty until we run the workflows the way they are intended to run.

True, but when it's an unfamiliar CI configuration, we can get a lot of that certainty working in a separate PoC/dev repo. We discussed on the call yesterday and I mentioned it in my review. There you can branch, fork, PR, merge without approval and check the results in a short feedback loop. Then once the configuration is known to be good, bring it back to the primary repo. We're not uncertain about the code or the build, just the CI process.

If it takes 2 PRs to figure out the CI process, setting up the sandbox is not worth the effort and I agree it makes sense to try a couple times. Now, if it takes 20 attempts, it was definitely worth the trouble to set up the sandbox. I grant that it's hard to know that ahead of time, so I asked on this 5th PR "How confident are we that the scans will be working after this?" and I didn't see an answer. 

I'll dismiss my review to unblock this effort, but (A) I don't see any value in leaving commented-out code here (that's why we have `git`) and (B) if the scans aren't succeessful after this PR, I will insist on removing the failing workflow until we see a PoC with it working.

PR #97, opened 2025-06-25
Remove resp.Diagnostics precondition to repair common Schema function

fincamd:
The resp.Diagnostics == nil check was breaking all refactored datasources since this pointer seems like it's nil until you add some diagnostics to it. Hence, we won't need this check in the common provider code.


dleehr (approved):

PabloHiro (approved):

AaronH88 (approved):
lgtm

PR #98, opened 2025-06-26
Remove checks that break Terraform RPC Flow

lranjbar:
These checks actually break the flow of Terraform's RPC calls. This is why we were having odd functionality using the base datasource.

The order of the checks are documented here: https://developer.hashicorp.com/terraform/plugin/framework/internals/rpcs#rpcs-and-framework-functionality


fincamd (approved):
I'll expect some resistance to remove all ctx and/or nil pointers checks. However, if these were breaking Terraform's normal execution flow it makes sense to remove them. Thanks for the PR @lranjbar.

LGTM!

PabloHiro (approved):

PR #99, opened 2025-06-30
Add Darwin and arm64 platform builds

davemulford:
Add Darwin and arm64 platform builds. Also fix archive format deprecation warning.


arrestle (approved):
LTGM

PR #102, opened 2025-07-01
Refactor NamedUrl Functions in preparation for Organization Data Source

lranjbar:
In prep for the new Organization data source this adds:

- Refactors `ReturnAAPNamedURL()` and extends it to new Named URL types.
- ~New `BaseDataSource` type `BaseDataSourceWithNamedUrl` for Organization / Inventory Source~ Now in #104 
- ~Moves base data source into its own file in prep for `BaseResource` creation. It's less confusing to have `BaseResource` and `BaseDataSource` in separate files because their SDKs are very close in name.~ 
 Now in #103

PR #104, opened 2025-07-01
Refactor Base API Models in preparation for Organization Data Source

lranjbar:
This PR aims to incorporate feedback from @fincamd and @PabloHiro on our API models.

Feedback comments:
- https://github.com/ansible/terraform-provider-aap/pull/102#discussion_r2177096948
- https://github.com/ansible/terraform-provider-aap/pull/81#discussion_r2144590495

Splitting out the API model changes from #102 so that we can review this separately.

Doing some research on the AWX API:
- All API calls have an `Id` and `URL`
- The next most common set is `Description`, `Name`, `Related`, `SummaryFields`, `Variables`. I put these in `BaseDetailAPIModelCommon` and `BaseDetailSourceModelCommon`
- Not all API calls have all of the above common things so I added more reusable models

For the provider:
- `APIModel` and `SourceModel` pairs are actually shared because `Resource` and `DataSource` in our provider so I renamed the `DataSourceModels` to `SourceModels`.

### DIFF
--- a/internal/provider/base_datasource_test.go
+++ b/internal/provider/base_datasource_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/datasourcevalidator"
-	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	fwdatasource "github.com/hashicorp/terraform-plugin-framework/datasource"
 	tfpath "github.com/hashicorp/terraform-plugin-framework/path"
 )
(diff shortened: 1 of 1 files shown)


fincamd (approved):
LGTM, but just one question. So Base models will include the necessary fields, but if there's an entity that doesn't have those common fields, the idea is then to compose it manually using the `BaseDetailAPIModel*` struct types? Like this:

```go
type MyCustomAAPEntity struct {
    BaseDetailAPIModelName
    BaseDetailAPIModelSummaryFields
}
```

We'd then need to override the set of methods of the datasource interface, right?

PR #105, opened 2025-07-03
Exclude _test.go files from sonarqube coverage

lranjbar:
Sonarqube is currently including _test.go files in coverage reports and it is skewing the results.


fincamd (approved):

arrestle (approved):
Excellent!

PR #107, opened 2025-07-09
AAP-46310 first draft replacement of pr#90

arrestle:


davemulford (changes requested):
Some minor name change requests. Looks good otherwise, great work!

davemulford (approved):
Implements the changes I asked about. Looks good!

internal/provider/base_datasource.go
```diff
@@ -207,11 +205,15 @@ func (d *BaseDataSource) ValidateConfig(ctx context.Context, req datasource.Vali
 		return
 	}
 
+	if IsValueProvidedOrPromised(data.Name) {
```
AaronH88, line 208: Same as above

internal/provider/base_datasource.go
```diff
@@ -207,11 +205,15 @@ func (d *BaseDataSource) ValidateConfig(ctx context.Context, req datasource.Vali
...
+		return
+	}
+
 	if !IsValueProvidedOrPromised(data.Id) {
 		resp.Diagnostics.AddAttributeWarning(
 			tfpath.Root("id"),
 			"Missing Attribute Configuration",
-			"Expected [id]",
+			"Expected [id] or [name]",
```
AaronH88, line 216: same as above

arrestle:
SonarQube does not bring in acceptance tests yet, so this  not correct yet.

Issue #76, opened 2025-06-04
aap_job error with validation due to unknown values


srlynch1:
Validation is not handling unknown values. Validation should be deferred in this scenario
The value is computed by datasource but is unknown at plan

```
│ Error: Missing Configuration for Required Attribute
│ 
│   with module.aap_postinstall.aap_job.config_vault_credentials,
│   on aap_postinstall/aap.tf line 28, in resource "aap_job" "config_vault_credentials":
│   28:   job_template_id = data.aap_job_template.vault_config.id
│ 
│ Must set a configuration value for the job_template_id attribute as the
│ provider has marked it as required.
│ 
│ Refer to the provider documentation or contact the provider developers for
│ additional information about configurable attributes that are required.
╵
```

tyraziel:
Hi @srlynch1 - are you able to provide any sample TF code to help us reproduce this error?  Thank you!

srlynch1:
https://github.com/Hashi-RedHat-APJ-Collab/platform-aap/tree/main

I have intentionally used explicit dependencies between modules, here.
This has a flow on impact though of causing unknown values in some scenarios which is fine and expected, but I generally expect the provider to not validate on unknown values

srlynch1:
The pattern here is unknown values are being treated as invalid when they should just return early. For provider framework guidance,  see 

https://developer.hashicorp.com/terraform/plugin/framework/validation

Reference below:

# 
**Note:** When implementing validation logic, configuration values may be [unknown](https://developer.hashicorp.com/terraform/plugin/framework/types#unknown) based on the source of the value. Implementations must account for this case, typically by returning early without returning new diagnostics.

PR #108, opened 2025-07-09
Fixes unknown values bug

### LINKED ISSUE #75: Inventory datasource returns error even though required attributes are set
Appears the real error is obfuscated in this scenario, as the required attributes are set
```
variable "job_template_name" {
  description = "Name of the AAP job template to run"
  type        = string
  default     = "Hashicorp Vault demo setup"
}

data "aap_job_template" "vault_config" {
  name = var.job_template_name
  organization_name = "Default"
}
```

https://github.com/ansible/terraform-provider-aap/blob/b8a81ac3b016da5f4c6131503c5864a63218677f/internal/provider/inventory_data_source.go#L108
```
module.aap_postinstall.aap_job.config_vault_credentials: Drift detected (update)
╷
│ Warning: Missing Attribute Configuration
│ 
│   with module.aap_postinstall.data.aap_job_template.vault_config,
│   on aap_postinstall/aap.tf line 22, in data "aap_job_template" "vault_config":
│   22: data "aap_job_template" "vault_config" {
│ 
│ Expected either [id] or [name + organization_name] pair
```

lranjbar:
Fixes #75 and #76
Closes AAP-47150 and AAP-47151

The confusion over this bug was starting to block progress on the Organization Data Source.

### DIFF
--- a/internal/provider/utils_test.go
+++ b/internal/provider/utils_test.go
@@ -127,7 +127,7 @@ func TestIsValueProvided(t *testing.T) {
 	}
 	for _, test := range testTable {
 		t.Run(test.testName, func(t *testing.T) {
-			actual := IsValueProvided(test.value)
+			actual := IsValueProvidedOrPromised(test.value)
 			if actual != test.expected {
 				t.Errorf("Expected %v but got %v", test.expected, actual)
 			}
@@ -135,6 +135,7 @@ func TestIsValueProvided(t *testing.T) {
 	}
 }
 
+// TODO: Replace ReturnAAPNamedURL with CreateNamedURL during Resource refactor
 func TestReturnAAPNamedURL(t *testing.T) {
 	var testTable = []struct {
 		testName    string
--- a/internal/provider/base_datasource.go
+++ b/internal/provider/base_datasource.go
@@ -203,11 +203,11 @@ func (d *BaseDataSource) ValidateConfig(ctx context.Context, req datasource.Vali
 		return
 	}
 
-	if IsValueProvided(data.Id) {
+	if IsValueProvidedOrPromised(data.Id) {
 		return
 	}
 
-	if !IsValueProvided(data.Id) {
+	if !IsValueProvidedOrPromised(data.Id) {
 		resp.Diagnostics.AddAttributeWarning(
 			tfpath.Root("id"),
 			"Missing Attribute Configuration",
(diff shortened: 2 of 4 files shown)


arrestle (changes requested):
This is excellent work! I feel you are spot-on with moving away from the IsValueProvided abstraction in favor of direct value checking (id.ValueInt64() != 0) in the new CreateNamedURL method. That is a real architectural improvement that makes the code cleaner and more maintainable (and I understand this required a lot of investigation on your part.)

I do have some difficulty with the IsValueProvided function because the name is confusing now. If a value is "Unknown" (a placeholder), how does that count as "Provided"? Would it be worth renaming this function to something clearer like ShouldProcessValue() or IsValueNotNull(), or perhaps deprecating it entirely since your new architecture handles this so much better?

[Unknown Values: The Secret to Terraform Plan](https://log.martinatkins.me/2021/06/14/terraform-plan-unknown-values/) explains that _Unknown_ values are the same _Known After Apply,_ in that they don't have the final values yet, but still might not be null.

dleehr:
> This is excellent work! I feel you are spot-on with moving away from the IsValueProvided abstraction in favor of direct value checking (id.ValueInt64() != 0) in the new CreateNamedURL method. That is a real architectural improvement that makes the code cleaner and more maintainable (and I understand this required a lot of investigation on your part.)
> 
> I do have some difficulty with the IsValueProvided function because the name is confusing now. If a value is "Unknown" (a placeholder), how does that count as "Provided"? Would it be worth renaming this function to something clearer like ShouldProcessValue() or IsValueNotNull(), or perhaps deprecating it entirely since your new architecture handles this so much better?
> 
> [Unknown Values: The Secret to Terraform Plan](https://log.martinatkins.me/2021/06/14/terraform-plan-unknown-values/) explains that _Unknown_ values are the same _Known After Apply,_ in that they don't have the final values yet, but still might not be null.

Thanks for the detailed review @arrestle. I'm looking through PRs this week and trying to see how I can help unblock things and bring attention where it's needed. So in this case, it sounds like you agree with the fix / code change, but your feedback is on the name of the `isValueProvided` function (which was not introduced in this PR).

Is that fair? Do you feel strongly that `isValueProvided` is an unsuitable name for this function that we must change right now? I do not want to dismiss your feedback (having been in your position many times), but I am hoping to understand the level of concern here.

internal/provider/utils_test.go
```diff
@@ -87,8 +88,56 @@ func checkBasicInventoryAttributes(
...
+	}{
+		{
+			testName: "value is int64(1)",
+			value:    types.Int64Value(1),
+			expected: true,
+		},
+		{
+			testName: "value is int64 unknown",
+			value:    types.Int64Unknown(),
+			expected: true,
```
arrestle, line 105: This illustrates my concern - the value is unknown but it's returning IsValueProvided = true

arrestle (changes requested):

arrestle (approved):

PR #110, opened 2025-07-10
README Updates

davemulford:
Adds text about linking the workflow template to the Default organization when running acceptance tests. Also updates the Supported Platforms section for the latest release.


dleehr:
Thanks for the docs updates!

arrestle (approved):

PR #118, opened 2025-07-21
Fix host resource update failing with retry logic due to consumed request body

arrestle:
## Problem

The host resource update operation was failing in step 3 of the acceptance test with HTTP 400 error: "name field is required".

## Root Cause

The issue was in the Update method's retry logic for handling HTTP 409 conflicts (when host is being used by running jobs). The `bytes.NewReader` was created once before the retry loop, but HTTP requests consume the reader's data. On subsequent retry attempts, the reader was already consumed (empty), so the API received an empty request body.

## Solution

Moved the `bytes.NewReader` creation inside the retry operation function so each retry attempt gets a fresh, complete request body.

## Before (broken):
```go
requestData := bytes.NewReader(updateRequestBody)  // Created once, reused
updateOperation := func() {
    return r.client.UpdateWithStatus(data.URL.ValueString(), requestData)  // Reusing consumed reader
}
```

## After (fixed):
```go
updateOperation := func() {
    requestData := bytes.NewReader(updateRequestBody)  // Fresh reader each time
    return r.client.UpdateWithStatus(data.URL.ValueString(), requestData)
}
```

## Testing

- ✅ `TestAccHostResource` now passes (was failing at step 3)
- ✅ All 3 test steps complete successfully  
- ✅ Test execution time: ~12 seconds

## Impact

This fix ensures that host updates work correctly when the host is temporarily unavailable due to running jobs, which is a legitimate use case in AAP environments.


dleehr (changes requested):
Hey @arrestle, thanks for marching forward on this. 

My main feedback is that the changes are about 900 lines, which is pretty far beyond the guidance we agreed on last week. So I will ask you to revisit this to deliver a smaller increment that the team can more readily review.

Related: while I haven't done a detailed review, my impression is that there are more changes here than what was asked in the story. If you have questions please ask and I'd be happy to set up some time.

dleehr:
I see this PR is still open and #120 has been opened to move forward. Can we close this PR so it's clear to the team that we're not expected to further review it?

PR #119, opened 2025-07-21
AAP-46310 [terraform-provider-aap] Organization Data Source - Failing 2.4 Acceptance Tests

arrestle:
## Fix optional description field in data sources

This PR makes the `description` field optional in base data source schemas to prevent test failures when entities don't have descriptions in AAP 2.4.

  ### Changes:
  - Updated `BaseDataSource` and `BaseDataSourceWithOrg` to mark description field as optional
  - Fixed README formatting for test setup instructions
  - Updated go.mod dependencies

  ### Background:
  AAP 2.4 entities may not always have descriptions, causing acceptance tests to fail when the description field was required. Making this field optional allows tests to pass while  still computing the description when present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,8 @@ The inventory resource test requires the AAP instance to have a second organizat
 ```bash
 export AAP_TEST_ORGANIZATION_ID=<the ID of Non-Default in your AAP instance>
 ```
-Create inventory `Inventory For Workflow"` on Default organization
-Create a Workflow Job Template called `Workflow with Inventory`
+1. Create inventory `Inventory For Workflow` on Default organization
+2. Create a Workflow Job Template called `Workflow with Inventory`
   - Assign organization to `Default`
   - Assign `Inventory For Workflow`
   - Make sure `Prompt on launch` **is not checked** for the inventory
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/hashicorp/terraform-plugin-framework-jsontypes v0.2.0
 	github.com/hashicorp/terraform-plugin-framework-validators v0.16.0
 	github.com/hashicorp/terraform-plugin-go v0.25.0
+	github.com/hashicorp/terraform-plugin-log v0.9.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.35.0
 	github.com/hashicorp/terraform-plugin-testing v1.6.0
 	github.com/stretchr/testify v1.8.4
@@ -47,7 +48,6 @@ require (
 	github.com/hashicorp/logutils v1.0.0 // indirect
 	github.com/hashicorp/terraform-exec v0.21.0 // indirect
 	github.com/hashicorp/terraform-json v0.23.0 // indirect
-	github.com/hashicorp/terraform-plugin-log v0.9.0 // indirect
 	github.com/hashicorp/terraform-registry-address v0.2.4 // indirect
 	github.com/hashicorp/terraform-svchost v0.1.1 // indirect
 	github.com/hashicorp/yamux v0.1.2 // indirect
--- a/internal/provider/base_datasource.go
+++ b/internal/provider/base_datasource.go
@@ -90,6 +90,7 @@ func (d *BaseDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, r
 				Description: fmt.Sprintf("Name of the %s", d.DescriptiveEntityName),
 			},
 			"description": schema.StringAttribute{
+				Optional:    true,
 				Computed:    true,
 				Description: fmt.Sprintf("Description of the %s", d.DescriptiveEntityName),
 			},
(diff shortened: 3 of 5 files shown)


dleehr (changes requested):
I don't think we should make this schema change. We should keep what we had (`Computed` without `Optional`) in the schema and either update the acceptance tests, the internal logic, or our setup steps to address the failing test. It's good to catch that an org's description behaves differently on 2.4 but changing the schema is not how to fix it.

`Optional` means that the user (practitioner in terraform's docs) can specify the value in their HCL (`.tf`) code but they don't have to. The problem is that for datasources, our provider doesn't use the `description` attribute, so it doesn't make sense to allow the user to provide it. These data sources consume `id` and `name` fields and we use those to make API calls to AAP to find the object they're referencing. `description` isn't part of that, so having it as an input of any kind (even `Optional`) in the schema for datasources is confusing.

The docs for StringAttribute https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/string#configurability explain the semantics:

> [`Optional`](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/string#optional) only: The value may be practitioner configured to a known value or null.
> [`Optional`](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/string#optional-1) and `Computed`: The value may be practitioner configured or the value may be set in provider logic when the practitioner configuration is null.

For our data sources, `description` should remain a **`Computed`** attribute. It's not something the user provides, it's always something we look up from AAP.

Sounds like for 2.5 it's populated as expected, but 2.4 behaves differently. A few follow-up questions then:

1. What's the specific difference with Organization in 2.4? Do organizations not have a description? Does `description` appear in the JSON but is `null`? Does the default organization (id=1) just not have the description filled out?
2. Earlier today you linked to a commit https://github.com/ansible/terraform-provider-aap/commit/9b42938da8f30d5497fe6b579504d59bf4a58040 that removed the checks for `description`. Does that succeed on both 2.4 and 2.5? I think that's still a valid test, it demonstrates that we can fetch an organization with an expected name.

internal/provider/base_datasource.go
```diff
@@ -90,6 +90,7 @@ func (d *BaseDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, r
 				Description: fmt.Sprintf("Name of the %s", d.DescriptiveEntityName),
 			},
 			"description": schema.StringAttribute{
+				Optional:    true,
```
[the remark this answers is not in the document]
arrestle: I tried that. If the value is empty it gets returned as a `nil`, which fails the `resource.TestCheckResourceAttrSet` test.

internal/provider/inventory_resource_test.go
```diff
@@ -176,6 +176,7 @@ func TestAccInventoryResource(t *testing.T) {
 	updatedDescription := "A test inventory"
 	updatedVariables := "{\"foo\": \"bar\"}"
 	resource.Test(t, resource.TestCase{
+
```
dleehr, line 179:
```suggestion
```

It's helpful to keep unnecessary changes out of a PR. This is a pretty minimal PR at this point (which is good!) so having a stray blank line stands out. I don't feel strongly about this specific line but I do ask to clean up unintended changes.

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/hashicorp/terraform-plugin-framework-jsontypes v0.2.0
 	github.com/hashicorp/terraform-plugin-framework-validators v0.16.0
 	github.com/hashicorp/terraform-plugin-go v0.25.0
+	github.com/hashicorp/terraform-plugin-log v0.9.0
```
dleehr, line 12: Similar question on this, yeah it's a trivial change (effectively just moving this line and removing the `// indirect` comment) but why is it here? Is someone's next commit going to move it back?

arrestle: `go mod tidy` did this.

README.md
```diff
@@ -67,13 +67,14 @@ The inventory resource test requires the AAP instance to have a second organizat
...
+1. Create inventory `Inventory For Workflow` on Default organization
+2. Create a Workflow Job Template called `Workflow with Inventory`
   - Assign organization to `Default`
   - Assign `Inventory For Workflow`
   - Make sure `Prompt on launch` **is not checked** for the inventory
   - Make sure `Prompt on launch` **is checked** for `Extra variables`
   - Add a default step and save
+3. If 2.4 set the description for Default Organization to `The default organization for Ansible Automation Platform`
 ```bash
 export AAP_TEST_WORKFLOW_INVENTORY_ID=<the ID of `Workflow with Inventory`>
```
tyraziel, line 79: Something needs to be added here to explain why we're exporting these additional environment variables.

arrestle: I agree but that is out of scope for this ticket. The only change we're making for this  is to set the description on the Default organization.

dleehr:
@arrestle I see you've requested reviews again from me and @lranjbar. We both approved yesterday, so there's no need for us to review/approve again.

The reason you cannot merge is that @tyraziel's review is requesting changes. So the next step is to work with @tyraziel to determine how to move forward.

tyraziel (approved):

PR #121, opened 2025-07-23
Fix issue in IsContextActive

davemulford:
Fixes an issue where checking if the current context is active would return the correct status while also writing erroneous error messages. This change also adds unit tests for the `IsContextActive` function and changes the function to pass the `diags` parameter by pointer.


AaronH88 (approved):
lgtm

dleehr (approved):

PR #123, opened 2025-07-30
[TRIVIAL] Fix aap_job using default inventory id

davemulford:
Prevents the aap_job resource from using the a default inventory-id of 1.

Fixes AAP-49876


arrestle (approved):
We have acceptance test https://github.com/ansible/terraform-provider-aap/pull/113/  and I believe we've uploaded the results of that on AAP-49876 so I think we're good here.

PR #127, opened 2025-08-04
Retry Host Deletion until Job Completes

lranjbar:
Issue #68 

This will attempt to retry to delete the host for a default of 30mins or until the job completion timeout has been reached.

Manual HCL testing:
```
resource "aap_host" "sleepyhost-test" {
  inventory_id = 1
  name         = "tf_host_foo"
  variables = jsonencode(
    {
      "ansible_host" : "127.0.0.1",
      "ansible_connection" : "local"
    }
  )
}

resource "aap_job" "sleepyjob" {
  job_template_id = <id_of_job_template_that_sleeps_for_15_seconds>
  inventory_id    = 1
}

terraform apply
terraform destroy -target=aap_host.sleepyhost-test
```

Watch for host deletion. The output should something like:

<details>

```
lranjbar@fedora:~/github/ansible/terraform-provider-aap$ terraform destroy -target=aap_host.sleepyhost-test
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - ansible/aap in /home/lranjbar/github/ansible/terraform-provider-aap
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become
│ incompatible with published releases.
╵
aap_host.sleepyhost-test: Refreshing state... [name=tf_host_foo]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  - destroy

Terraform will perform the following actions:

  # aap_host.sleepyhost-test will be destroyed
  - resource "aap_host" "sleepyhost-test" {
      - enabled      = true -> null
      - id           = 91 -> null
      - inventory_id = 1 -> null
      - name         = "tf_host_foo" -> null
      - url          = "/api/controller/v2/hosts/91/" -> null
      - variables    = jsonencode(
            {
              - ansible_connection = "local"
              - ansible_host       = "127.0.0.1"
            }
        ) -> null
    }

Plan: 0 to add, 0 to change, 1 to destroy.
╷
│ Warning: Resource targeting is in effect
│ 
│ You are creating a plan with the -target option, which means that the result of this plan may not represent all of the changes
│ requested by the current configuration.
│ 
│ The -target option is not for routine use, and is provided only for exceptional situations such as recovering from errors or
│ mistakes, or when Terraform specifically suggests to use it as part of an error message.
╵

Do you really want to destroy all resources?
  Terraform will destroy all your managed infrastructure, as shown above.
[...]

PR #129, opened 2025-08-06
[TRIVIAL] Add retry mocks and update generate docs to use go install

lranjbar:
- Adds retry mocks from #127.
- Adds retry unit tests from #127 and uses t.Skip() 
- Adds Retry function signatures from #127, without implementing them.
- Updates generate docs CI job to use go install
- Clean up some formatting in client_test.go


arrestle (approved):
Looks good!

PR #132, opened 2025-08-07
[AAP-47221] 2 - Fix logging

arrestle:

Issue #78, opened 2025-06-04
aap_job - completion behaviour


srlynch1:
The AAP provider appears to be logging to stdout which is throwing warning errors and unexpected data in logs. tflog should be used for logging rather than just writing to stdout

Additionally whilst the job polls and waits for completion on successful completion the job status is not updated so the output doesn't reflect the current status

Expectation is the [status](https://registry.terraform.io/providers/ansible/aap/latest/docs/resources/job#status-1) (String) Status of the job will be updated to reflect the current job status its still currently outputting pending when its in a successful state

```
module.aap_postinstall.aap_job.config_vault_credentials: Modifying...
2025-06-04T23:07:28.183Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: pending"
2025-06-04T23:07:29.090Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: pending"
2025-06-04T23:07:30.490Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: pending"
2025-06-04T23:07:32.902Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: pending"
module.aap_postinstall.aap_job.config_vault_credentials: Still modifying... [10s elapsed]
2025-06-04T23:07:37.301Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: pending"
2025-06-04T23:07:45.722Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: running"
module.aap_postinstall.aap_job.config_vault_credentials: Still modifying... [20s elapsed]
2025-06-04T23:07:56.125Z [WARN]  unexpected data: registry.terraform.io/ansible/aap:stdout="Job ID: 13, Current Status: successful"
module.aap_postinstall.aap_job.config_vault_credentials: Modifications complete after 29s

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.

Outputs:
aap_job_status = "pending"
```

tyraziel:
Fixed in 1.3.0 - https://github.com/ansible/terraform-provider-aap/releases/tag/v1.3.0

PR #133, opened 2025-08-07
[TRIVIAL] [AAP-47221] 3 - Failing acceptance test (To be fixed by PR 1 & 2)

arrestle:
### Summary [AAP-47221] - Issue #78
Adds acceptance test that demonstrates the job status bug where wait_for_completion=true jobs get stuck in pending status instead of updating to final states.
Test behavior:

1.  PASSES with PR #131 applied (job reaches final state)
2.  FAILS without fixes (job stuck in pending, times out)

Changes:

- New TestAccAAPJob_WaitForCompletion test with custom validation
- Uses sleepy job template via AAP_TEST_JOB_FOR_HOST_RETRY_ID environment variable
- Validates any final state (successful, failed, error, canceled) instead of hardcoding one status

Dependencies:

1. Requires PR #131 (diagnostics error handling)
2. Requires PR #132 (logging fixes)

### DIFF
--- a/internal/provider/job_resource_test.go
+++ b/internal/provider/job_resource_test.go
@@ -415,6 +415,40 @@ func TestAccAAPJob_UpdateWithTrigger(t *testing.T) {
 	})
 }
 
+// TestAccAAPJob_WaitForCompletion tests that job status is correctly updated to final state
+// when wait_for_completion=true. This test demonstrates the bug described in AAP-47221.
+// Expected to FAIL on main branch, PASS after PR #131 and #132 are merged.
+func TestAccAAPJob_WaitForCompletion(t *testing.T) {
+	jobTemplateID := os.Getenv("AAP_TEST_JOB_TEMPLATE_ID")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccJobResourcePreCheck(t) },
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccJobWithWaitForCompletion(jobTemplateID),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckJobExists,
+					// This check should FAIL on main branch due to AAP-47221 bug
+					// The job status should be "successful" or "failed", not "pending"
+					resource.TestCheckResourceAttrWith("aap_job.test", "status", func(value string) error {
+						if value == "pending" {
+							return fmt.Errorf("AAP-47221 bug: job status is 'pending' instead of final state")
+						}
+						if !IsFinalStateAAPJob(value) {
+							return fmt.Errorf("job status '%s' is not a final state", value)
+						}
+						return nil
+					}),
+					// Verify wait_for_completion was actually used
+					resource.TestCheckResourceAttr("aap_job.test", "wait_for_completion", "true"),
+					resource.TestCheckResourceAttr("aap_job.test", "wait_for_completion_timeout_seconds", "300"),
+				),
+			},
+		},
+	})
+}
+
 // testAccCheckJobPause is designed to force the acceptance test framework to wait
 // until a job is finished. This is needed when the associated inventory also must be
 // deleted.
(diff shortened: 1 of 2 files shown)


arrestle:
[PR-133.failing.testacc.log](https://github.com/user-attachments/files/21740791/PR-133.failing.testacc.log)
[PR-133.passing.testacc.log](https://github.com/user-attachments/files/21740792/PR-133.passing.testacc.log) Manual merging of changes required from the PR #1 and PR #2. We should retest before merging.

internal/provider/job_resource_test.go
```diff
@@ -407,6 +408,40 @@ func TestAccAAPJob_UpdateWithTrigger(t *testing.T) {
...
+					// The job status should be "successful" or "failed", not "pending"
+					resource.TestCheckResourceAttrWith("aap_job.test", "status", func(value string) error {
+						if value == "pending" {
+							return fmt.Errorf("AAP-47221 bug: job status is 'pending' instead of final state")
+						}
+						if !IsFinalStateAAPJob(value) {
+							return fmt.Errorf("job status '%s' is not a final state", value)
+						}
+						return nil
+					}),
```
dleehr, line 435: This is pretty cool that the check logic can be a function, and I can see that it's useful while we're developing.

Given that this test should only pass once we have implemented the fix, what do you think about simplifying it to just check if the job status  is `successful`?

dleehr: I'm asking what **you** think 😄.

I pointed out that it looks to me like this could be simplified this to a single line, but I wanted your opinion . If you agree that would be better, change it. If you disagree, that's fine too.

arrestle: @dleehr I prefer keeping the custom validation:
Better error messages - when it fails, you immediately know it's the AAP-47221 bug instead of just "expected successful, got pending"
More realistic - jobs can end in failed, error, or canceled too, not just successful
The custom function makes it clear we're testing the core issue: jobs getting stuck in pending instead of reaching any final state.

internal/provider/job_resource_test.go
```diff
@@ -415,6 +417,40 @@ func TestAccAAPJob_UpdateWithTrigger(t *testing.T) {
 	})
 }
 
+// TestAccAAPJob_WaitForCompletion tests that job status is correctly updated to final state
+// when wait_for_completion=true. This test demonstrates the bug described in AAP-47221.
```
davemulford, line 421: Need team consensus on this, but we may want to remove references to internal Jira issues and other PRs. Not everyone will have access to them?

internal/provider/job_resource_test.go
```diff
@@ -415,6 +417,40 @@ func TestAccAAPJob_UpdateWithTrigger(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccJobWithWaitForCompletion(jobTemplateID),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckJobExists,
+					// This check should FAIL on main branch due to AAP-47221 bug
+					// The job status should be "successful" or "failed", not "pending"
+					resource.TestCheckResourceAttrWith("aap_job.test", "status", func(value string) error {
+						if value == statusPendingConst {
+							return fmt.Errorf("AAP-47221 bug: job status is 'pending' instead of final state")
```
davemulford, line 438: Same as above comment.

internal/provider/job_resource_test.go
```diff
@@ -415,6 +417,40 @@ func TestAccAAPJob_UpdateWithTrigger(t *testing.T) {
...
+						}
+						if !IsFinalStateAAPJob(value) {
```
davemulford, line 440: Is the above `value == statusPendingConst` necessary if what we really care about is that the job is in a "final" state?

davemulford (changes requested):
Requesting some changes based on team consensus about whether we should reference Jira issues and PRs in comments and log messages.

internal/provider/job_resource_test.go
arrestle: I've asked about this before and because it's very difficult to map a PR back to the requested work we were directed to include Jira numbers (but not links) on PRs.

arrestle: removed jira reference and replaced with issue.

internal/provider/job_resource_test.go
arrestle: removed.

davemulford (approved):
LGTM

PR #140, opened 2025-09-15
Add 403 Forbidden as a retryable error

davemulford:
This change makes the 403 Forbidden HTTP response an error condition that can be retried. While this is not an error response that is normally retried, the thought here is that AAP 2.4 is sending an incorrect response code.

FIXES: AAP-51627

### DIFF
--- a/internal/provider/retry.go
+++ b/internal/provider/retry.go
@@ -68,9 +68,14 @@ var (
 	// - HTTP 502: Bad gateway
 	// - HTTP 503: Service unavailable
 	// - HTTP 504: Gateway timeout
+	//
+	// HTTP 403: Forbidden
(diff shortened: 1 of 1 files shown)


dleehr (commented):
> While this is not an error response that is normally retried, the thought here is that AAP 2.4 is sending an incorrect response code.

I agree that 403 is not something I would expect to be recoverable through retries, so I appreciate the additional comments here. That's what I like to see whenever we're doing something unexpected :) 

My only question before approving - Does this fix the flakiness? I see from the ticket that we'd sometimes get a 403, and re-running the tests eventually worked. So now that we're retrying on 403, does AAP eventually respond with a different status code?

Is that a bug we can report to controller (or reference if one exists)?

PR #147, opened 2025-09-22
feat: Add support for actions with job templates

danquack:
#146 is meant to precede this PR. This PR will seem huge otherwise.

```
go test -v -run TestAccAAPJobAction_basic ./...
?       github.com/ansible/terraform-provider-aap       [no test files]
=== RUN   TestAccAAPJobAction_basic
    provider_test.go:40: 'AAP_TOKEN' is set, using token authentication in acceptance tests
--- PASS: TestAccAAPJobAction_basic (11.66s)
PASS
ok      github.com/ansible/terraform-provider-aap/internal/provider     11.670s
```


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM hashicorp/terraform:1.14.0-beta as terraform
```
danquack, line 1: This is just so you can test it without having to download a beta provider.

internal/provider/job_action_test.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			// Parse the job creation time (assuming RFC3339 format from API)
+			jobTime, err := time.Parse(time.RFC3339, createdStr)
+			if err != nil {
+				continue
+			}
+
+			// Check if job was created within 5 minutes of the target timestamp
+			timeDiff := jobTime.Sub(targetTime).Abs()
```
danquack, line 101: Id be open to thoughts on if there's a better way to determine if an action, which isn't written to state kicks off. I was able to see the action start in the ui, but I could see where maybe this produces a false positive if another job is launched with this template id.

PR #150, opened 2025-09-24
[AAP-52742] 2: EDA Event Stream Data Source Implementation

davemulford:
This implements the EDA Event Stream Data Source that allows for discovering the event stream URL for use with TF Actions.

Depends on #144
Implements AAP-52742

### DIFF
--- a/internal/provider/base_eda_datasource.go
+++ b/internal/provider/base_eda_datasource.go
@@ -99,7 +99,15 @@ func (d *BaseEdaDataSource) Read(ctx context.Context, req datasource.ReadRequest
 	resp.Diagnostics.Append(req.Config.Get(ctx, &state)...)
 
 	// Create the EDA path with query parameters
-	resourceURL := path.Join(d.client.getEdaApiEndpoint(), d.ApiEntitySlug)
+	edaEndpoint := d.client.getEdaApiEndpoint()
+	if edaEndpoint == "" {
+		resp.Diagnostics.AddError(
+			"EDA API Endpoint is empty",
+			"Expected a valid endpoint but was an empty string. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	resourceURL := path.Join(edaEndpoint, d.ApiEntitySlug)
 	params := map[string]string{
 		"name": state.Name.ValueString(),
 	}


internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+			},
+			"url": schema.StringAttribute{
+				Computed: true,
+			},
+			"id": schema.Int64Attribute{
+				Computed: true,
+			},
+		},
+	}
+}
```
dleehr, line 56: Since `func Schema()` is implemented here in `BaseEdaDataSource` (and not on the concrete `EDAEventStreamDataSource` struct), that suggests we expect every `datasource` that we will implement under EDA to have these attributes `name`, `url`, and `id`. Is that the intent?

internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+// ConfigValidators validates configuration in a declarative way.
+func (d *BaseEdaDataSource) ConfigValidators(_ context.Context) []datasource.ConfigValidator {
+	// You have at least an name
+	return []datasource.ConfigValidator{
+		datasourcevalidator.Any(
+			datasourcevalidator.AtLeastOneOf(
+				tfpath.MatchRoot("name")),
+		),
+	}
+}
```
dleehr, line 67: Since `name` is already a required attribute in the schema, is this validator necessary? 

I think in general validators can be used for complex either/or relationships or to make sure strings match some pattern. Maybe we have these in other datasources, but I think if we're just making sure `name` exists, the `Required: true` on the Attribute handles it.

internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+
+// ValidateConfig validates configuration in an imperative way.
+func (d *BaseEdaDataSource) ValidateConfig(ctx context.Context, req datasource.ValidateConfigRequest, resp *datasource.ValidateConfigResponse) {
```
dleehr, line 70: Do we need both `ValidateConfig` and `ConfigValidators`?

My understanding from https://developer.hashicorp.com/terraform/plugin/framework/data-sources/validate-configuration is that `ConfigValidators` is the declarative way and is probably less code, but if you need to write imperative logic to check things, implement `ValidateConfig`.

internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+		return
+	}
+
+	if !IsValueProvidedOrPromised(data.Name) {
+		resp.Diagnostics.AddAttributeWarning(
+			tfpath.Root("id"),
+			"Missing Attribute Configuration",
+			"Expected [id]",
+		)
+	}
```
dleehr, line 100: I had a couple questions but first let me share my understanding of this code.

It looks to me like we first read a `BaseEdaSourceModel` from `req.Config` (no confusion there). Then, we check if that data has a **Name** (`IsValueProvidedOrPromised(data.Name)`).

If it does, 👍 . If not, we add a warning about a missing **id** (not name) 🤔.

So my first question is - why do we need to do that? Does the schema not handle this? Second question - if we do, should this be `name` and not `id`?

As an exercise, I'd be interested to see the output of `terraform validate` on the following snippet with all of these validators removed:

```hcl
data.aap_eda_eventstream {
  # empty on purpose
}
```

internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Check that the current context is active
+	if !IsContextActive("Configure", ctx, &resp.Diagnostics) {
+		return
+	}
+
+	// Check that the provider data is configured
+	if req.ProviderData == nil {
+		return
```
dleehr, line 118: Since `req.ProviderData` is what provides us the `AAPClient`, if we don't have it we should add an error to the diagnostics.

internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+	responseBody, diags := d.client.GetWithParams(resourceURL, params)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
dleehr, line 160: Looks like duplicate code here, we should only need to do this once.

internal/provider/base_eda_datasource.go
```diff
@@ -0,0 +1,198 @@
...
+	// Check Read preconditions
+	if !DoReadPreconditionsMeet(ctx, resp, d.client) {
+		return
+	}
+
+	// Read Terraform configuration data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &state)...)
+
+	// Create the EDA path with query parameters
+	resourceURL := path.Join(d.client.getEdaApiEndpoint(), d.ApiEntitySlug)
```
dleehr, line 147: What happens if the EDA is not present? I think `d.client.getEdaApiEndpoint()` will return an empty string and this will probably lead to a 404.

Maybe we could check if `d.client.getEdaApiEndpoint()` is empty and add an error to `resp.Diagnostics` noting that the EDA API is unavailable?

internal/provider/client.go
```diff
@@ -68,14 +74,46 @@ func readApiEndpoint(client ProviderHTTPClient) (string, diag.Diagnostics) {
...
+		err = json.Unmarshal(body, &response)
+		if err != nil {
+			diags.AddError(
+				fmt.Sprintf("Unable to parse EDA API endpoint response: %s", string(body)),
+				fmt.Sprintf("Unexpected error: %s", err.Error()),
+			)
+			return "", "", diags
+		}
+
+		// URL Parse because the EDA endpoint's current_version gives a full URL instead of a URL path
```
dleehr, line 102: Interesting. So the format of `current_version` from controller is not consistent with EDA? I notice we don't need to `url.Parse()` that response from controller

internal/provider/general_api_models.go
```diff
@@ -36,6 +36,16 @@ type BaseDetailAPIModelWithOrg struct {
 	Organization  int64                 `json:"organization"`
 }
 
+type BaseEdaAPIModel struct {
+	Name string `json:"name"`
+	Id   int64  `json:"id"`
+	URL  string `json:"url"`
+}
```
dleehr, line 43: Similar feedback to the `Schema` in base_eda_datasource.go. Do we expect everything under the EDA API to have these fields?

dleehr (changes requested):
Very exciting to see this progress! I'm specifically requesting changes around the validators in the datasource, since I think there's some duplication there (and maybe we don't need that).

Otherwise it looks great. I left a few minor suggestions here and there and raised a couple questions about the abstraction into the base and concrete structs - That's just a conversation starter, not something for which I'm requesting any specific change. We can always refactor that later if needed.

Oh and it looks like CI found a couple of things in a Mock that need to change to align with the Client changes.

Regarding adding tests - is that planned for a future PR?

internal/provider/client.go
davemulford: Yes I was surprised to learn they have different behavior.

Here is what the controller endpoint response looks like:
```json
{
  "description": "AWX REST API",
  "current_version": "/api/controller/v2/",
  "available_versions": {
    "v2": "/api/controller/v2/"
  },
  "custom_logo": "",
  "custom_login_info": "",
  "login_redirect_override": ""
}
```

Here is what the EDA endpoint response looks like:
```json
{
  "current_version": "http://localhost:44925/api/eda/v1/",
  "available_versions": {
    "v1": "http://localhost:44925/api/eda/v1/"
  }
}
```

internal/provider/base_eda_datasource.go
davemulford: I had looked at a few of the object definitions in the Swagger docs and the ones I looked at (Resource, ResourceType, RoleDefinition) all have those attributes. Looking a second time at other objects (ActivationInstance and Rulebook) they have the `id` and `name`, but not the `url`.

I think this method can be moved into the `EDAEventStreamDataSource` definition as you mention.

internal/provider/base_eda_datasource.go
davemulford: Agreed, see above comment about removing both `ConfigValidators` and `ValidateConfig`.

internal/provider/base_eda_datasource.go
davemulford: Good catch! This was a copy/paste error and I didn't update the warning text. I'll make sure I do that.

internal/provider/base_eda_datasource.go
davemulford: I'll add this in the next update.

internal/provider/base_eda_datasource.go
davemulford: Good point. There is nothing preventing someone from attempting to use the provider against an AAP deployment that does not have EDA set up. I'll add a check and error out if the endpoint is empty.

davemulford:
Thanks @dleehr for the review!

> Very exciting to see this progress! I'm specifically requesting changes around the validators in the datasource, since I think there's some duplication there (and maybe we don't need that).

I've replied inline to each comment thread. I like all the changes you've mentioned, so I plan to do those in this PR instead of having to refactor later.

> Oh and it looks like CI found a couple of things in a Mock that need to change to align with the Client changes.

I totally blanked on this before creating the PR. I've fixed them locally and will push changes to fix this right after I post this comment.

> Regarding adding tests - is that planned for a future PR?

I was going to add them in this PR, unless you think that would make it too large?

internal/provider/base_eda_datasource.go
davemulford: @dleehr, adding an error to the diagnostics here will always result in an error shown by the client. This is due to Terraform calling the data source `Configure` function multiple times. I've added a few logging statements to the `Configure` function as a test.

Here are the places I've added logging to show which path the code takes:
```diff
+	tflog.Error(ctx, "BaseEdaDataSource: before checking req.ProviderData")
	// Check that the provider data is configured
	if req.ProviderData == nil {
+		tflog.Error(ctx, "BaseEdaDataSource.Configure: ProviderData was nil")
		return
	}

	client, ok := req.ProviderData.(*AAPClient)
	if !ok {
		resp.Diagnostics.AddError(
			"Unexpected Data Source Configure Type",
			fmt.Sprintf("Expected *AAPClient, got: %T. Please report this issue to the provider developers.", req.ProviderData),
		)
		return
	}

+	tflog.Error(ctx, "BaseEdaDataSource: setting client")
	d.client = client
```

This is a snippet of the log messages when running `terraform apply`:
```
2025-09-25T09:06:00.886-0400 [ERROR] provider.terraform-provider-aap: BaseEdaDataSource: before checking req.ProviderData: ...
2025-09-25T09:06:00.886-0400 [ERROR] provider.terraform-provider-aap: BaseEdaDataSource.Configure: ProviderData was nil: ...
2025-09-25T09:06:01.267-0400 [ERROR] provider.terraform-provider-aap: BaseEdaDataSource: before checking req.ProviderData: ...
2025-09-25T09:06:01.267-0400 [ERROR] provider.terraform-provider-aap: BaseEdaDataSource: setting client: ...
2025-09-25T09:06:01.268-0400 [ERROR] provider.terraform-provider-aap: BaseEdaDataSource: before checking req.ProviderData: ...
2025-09-25T09:06:01.268-0400 [ERROR] provider.terraform-provider-aap: BaseEdaDataSource: setting client: ...
```

Hashicorp's [data source example](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-data-source-read#implement-data-source-client-functionality) shows a nil check with a return and no error added to the diagnostics. I'm not sure if that's how it's intended to be done in production-ready providers, or if that's just sample code.

Is there something you know of that would allow us to add an error here, without always erroring?

internal/provider/base_eda_datasource.go
davemulford: I've switched to using the same pattern used by the BaseResource struct, where a `GetBaseAttributes` function allows descendant structs to use the base attributes and add its own, while still providing a base implementation of the `Schema` function.

internal/provider/base_eda_datasource.go
davemulford: I've removed the `ConfigValidators` and `ValidateConfig` functions, which also removed this code.

internal/provider/base_eda_datasource.go
davemulford: I've added the check for empty string!

dleehr (approved):
> Very exciting to see this progress! I'm specifically requesting changes around the validators in the datasource, since I think there's some duplication there (and maybe we don't need that).

Thanks for cleaning up the validation logic, I think this is simpler and a better fit, so 👍 . I've looked through the other changes and it makes sense. I do think we still have a bit of a disconnect on the attributes/fields that are in the base implementation vs the concrete datasource. For example, the base structs that define the API and terraform state models include the url field, but this is excluded from the base schema. Maybe this is the existing pattern in the other bases, I think we'll just address that in the future when we have more data sources in EDA.

This looks like a good increment 👍. Test coverage is low, but the current size is manageable so I'm OK moving forward. I trust that there will be additional PRs to address the testing gap.

dleehr:
> Testing will be created in later PRs and is tracked by https://issues.redhat.com/browse/AAP-54793.

Thanks for the info. FYI for other reviewers, the standard here is to review the code with a priority to deliver the dev preview release. We are acknowledging a gap in test coverage and committing to deliver that in our next iteration.

So please review with that in mind.

tyraziel (approved):

PR #156, opened 2025-10-06
[AAP-54793] 1: Add client readApiEndpoint tests

davemulford:
This adds a number of test cases to work out the various code paths in the `readApiEndpoint()` function. The tests work by creating multiple http servers that respond with various values for the `/api/`, `/api/controller`, and `/api/eda` endpoints.

Related: AAP-54793

### DIFF
--- a/internal/provider/client_test.go
+++ b/internal/provider/client_test.go
@@ -19,6 +19,14 @@ const (
 type MockAuthenticator struct {
 }
 
+type readApiEndpointTestCase struct {
+	name                   string
+	url                    string
+	expectedControllerPath string
+	expectedEDAPath        string
+	diagsShouldHaveErr     bool
+}
+
 func (m *MockAuthenticator) Configure(_ *http.Request) {
 	// Do nothing
 }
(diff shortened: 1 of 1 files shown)


internal/provider/client_test.go
```diff
@@ -70,25 +84,201 @@ func TestReadApiEndpoint(t *testing.T) {
...
-			assert.Equal(t, tc.expectedEDAPath, client.getEdaApiEndpoint())
+		t.Run(tc.name, func(t *testing.T) {
+			executeReadApiEndpointTestCase(t, tc)
 		})
 	}
 }
 
+func executeReadApiEndpointTestCase(t *testing.T, tc readApiEndpointTestCase) {
+	// readApiEndpoint() is called when creating client
+	client, diags := NewClient(tc.url, &MockAuthenticator{}, true, 0)
```
dleehr, line 267: I see that you moved this check to its own helper function, rather than having it inside the `t.Run()` block. No objection there (since it seems to be de-duplicating common code), but I think when you do this you should call [t.Helper()](https://pkg.go.dev/testing#T.Helper). See also https://dev.to/eminetto/using-test-helpers-in-go-1o55.

dleehr (changes requested):
Additional tests look good and appreciate the cleanup/refactoring to use constants and deduplicate code.

Requesting changes just for a missing call to `t.Helper()`, otherwise LGTM

dleehr (approved):

matoval (approved):

internal/provider/client_test.go
```diff
@@ -10,9 +10,23 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	apiEndpoint        = "/api/"
+	controllerEndpoint = "/api/controller/"
+	edaEndpoint        = "/api/eda/"
```
PabloHiro, line 16:
```suggestion
	apiEndpoint        = "/api/"
	controllerEndpoint = "/api/controller/"  // Base API endpoint for Controller in AAP >= 2.5
	edaEndpoint        = "/api/eda/"  // Base API endpoint for EDA in AAP >= 2.5
```

internal/provider/client_test.go
```diff
@@ -70,25 +84,202 @@ func TestReadApiEndpoint(t *testing.T) {
...
-			assert.Equal(t, false, diags.HasError(), fmt.Sprintf("readApiEndpoint() returns errors (%v)", diags))
-			assert.Equal(t, tc.expectedControllerPath, client.getApiEndpoint())
-			assert.Equal(t, tc.expectedEDAPath, client.getEdaApiEndpoint())
+		t.Run(tc.name, func(t *testing.T) {
+			executeReadApiEndpointTestCase(t, tc)
 		})
 	}
 }
 
+func executeReadApiEndpointTestCase(t testing.TB, tc readApiEndpointTestCase) {
```
PabloHiro, line 265: Nit: Can we place this before the tests themselves? I think it improves readability. One starts reading at the beginning of this file, sees this, and then the tests are easy to understand.

Right now I need to scroll down, read it, and then scroll up to continue reading the tests

PabloHiro (changes requested):

internal/provider/client_test.go
```diff
@@ -70,25 +84,202 @@ func TestReadApiEndpoint(t *testing.T) {
...
+
+func TestReadApiEndpointForEDA(t *testing.T) {
+	serverWithMissingEDAEndpoint := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case apiEndpoint:
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{"apis":{"gateway": "/api/gateway/", "controller": "/api/controller/", "eda": "/api/eda/"}}`)) //nolint:errcheck
+		case controllerEndpoint:
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{"current_version": "/api/controller/v2/"}`)) //nolint:errcheck
```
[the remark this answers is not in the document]
davemulford: It's needed here to allow `readApiEndpoint()` to reach the section of code that attempts to discover the EDA endpoint. See the function flow below.

https://github.com/ansible/terraform-provider-aap/blob/f606a3ad86b4a40e0d0ce4e9b3ad45fb4e0bc9ff/internal/provider/client.go#L70-L94

PabloHiro (approved):

PR #168, opened 2025-11-17
Rebase of dev-preview on main

PabloHiro:
AAP-57267

Validation progress:

- [x] AAP 2.4
- [x] AAP 2.5
- [x] AAP 2.6

UPDATE: Acceptance tests passing against all three versions of AAP.


dleehr:
@tyraziel on the documentation point, is something missing that we should address ahead of the next release? If you spotted a gap, please share the feedback!

PR #191, opened 2026-05-31
Remove host from state on 404

### LINKED ISSUE #189: aap_host Read errors on 404 instead of removing from state — a missing host wedges plan/apply AND destroy (same bug as #59, fixed for aap_job in #61)
### Summary

`aap_host`'s `Read` method returns a hard error when the AAP API responds `404` for a host that is recorded in Terraform state but no longer exists in AAP. Instead of treating the `404` as "this resource is gone" and removing it from state (so the next plan recreates it), the provider surfaces the raw HTTP error. Because `Read` runs during refresh, this blocks `plan`, `apply`, **and `destroy`** — the workspace cannot be reconciled or even torn down without a manual `terraform state rm`.

This is the same failure mode and root cause as #59 / #139 (`aap_job` / workflow jobs), fixed in #61 / #153 — but that fix only updated `job_resource.go`; `host_resource.go`'s `Read` still surfaces the error. Reproduced on v1.5.0.

### Provider / Terraform versions

- `ansible/aap` v1.5.0
- Terraform v1.15.x (HCP Terraform, agent execution mode)

### Error

```
Error: Unexpected HTTP status code received for GET request to path
https://<aap-controller>/api/controller/v2/hosts/249/
Expected one of ([200]), got (404). Response details: No Host matches the given query.
  with aap_host.vm_hosts["host-01"]
  on inventory.tf line 32, in resource "aap_host" "vm_hosts"
```

### What happened (observed sequence)

The host was **not** deleted out-of-band — it disappeared as a side effect of the same `terraform apply`. Hosts were placed into groups via `aap_group` + `aap_host.groups`. A configuration change left one `aap_group` unused, so the plan deleted that group and updated the host's group membership. The observed apply order was:

1. create the new `aap_group`
2. **delete** the now-unused `aap_group`
3. update the `aap_host` → `PUT /api/controller/v2/hosts/249/` → **404**

After step 2 the host no longer existed in AAP, so the host update failed — and from then on **every** `plan` / `apply` / `destroy` fails at refresh with the `GET .../hosts/249/ → 404` shown above.

Whether deleting an `aap_group` should remove its member host is a separate question; regardless of *how* the host came to be missing, the bug this issue is about is that `aap_host.Read` should tolerate the `404` rather than erroring.

### Expected behaviour

On a `404`, `aap_host.Read` should remove the resource from state (e.g. `resp.State.RemoveResource(ctx)`) so the next plan marks it as disappeared and recreates it — exactly as #61 did for `aap_job`:

> "we remove the resource from the state if a 404 is encountered, so that the plan marks it as disappeared"

### Suggested fix

Apply the same `404` → remove-from-state handling in `internal/provider/host_resource.go`'s `Read`. The same gap very likely affects `aap_group` and `aap_inventory` `Read` as well — worth covering all three so a missing managed object doesn't wedge the workspace (including `destroy`).

### Workaround

```
terraform state rm 'aap_host.vm_hosts["<key>"]'
```

then `terraform apply` recreates the host.

yzhkali:
## Summary
- handle `aap_host` read 404s the same way job resources do
- remove the host from Terraform state so the next plan can recreate it instead of blocking refresh/destroy
- add a regression test for the 404 read path

Closes #189

## Tests
- `go test ./internal/provider -run TestHostResourceReadRemovesStateOnNotFound`
- `go test ./internal/provider`

## Bug Fixes
* Host resources are now automatically removed from state when they cannot be accessed
* Warning notifications alert users when host resources are unavailable

### DIFF
--- a/internal/provider/host_resource.go
+++ b/internal/provider/host_resource.go
@@ -226,7 +226,15 @@ func (r *HostResource) Read(ctx context.Context, req resource.ReadRequest, resp 
 	}
 
 	// Get latest host data from AAP
-	readResponseBody, diags := r.client.Get(data.URL.ValueString())
+	readResponseBody, diags, status := r.client.GetWithStatus(data.URL.ValueString(), nil)
+	if status == http.StatusNotFound {
+		resp.Diagnostics.AddWarning(
+			"Host not found",
+			"The host was not found. It may have been deleted. The host will be recreated.",
+		)
+		resp.State.RemoveResource(ctx)
+		return
+	}
 	resp.Diagnostics.Append(diags...)
 	if resp.Diagnostics.HasError() {
 		return
--- a/internal/provider/host_resource_test.go
+++ b/internal/provider/host_resource_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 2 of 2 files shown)


yzhkali:
Closing for now after prolonged inactivity and no maintainer review since May 31. GitHub reports the branch as mergeable but blocked, while the only automated review was skipped. Linked issue #189 remains open; the change can be refreshed against current main if the project confirms it still wants this 404 state-removal behavior.

